PR #1, opened 2023-02-27
Conanfile.txt to conanfile.py

RubenRBS:
This comes from @memsharded's work in https://github.com/conan-io/conan/pull/13218, cleaned up 
a bit.

The documentation suggests to independently clone this repo and then run `conan config install` because installing a whole repo from its URL copies _every_ file, and I don't think the users would be too happy having our supporting files in their Conan HOME (Such as the README)


memsharded:
> The documentation suggests to independently clone this repo and then run conan config install because installing a whole repo from its URL copies every file, and I don't think the users would be too happy having our supporting files in their Conan HOME (Such as the README)

I agree, and I think this deserves a small feature, that is a ``.conanignore`` file in ``config install`` repos that allow installing from the root URL and not copying all the non-related files. I am thinking for example, that we also want to add tests to this repo, but we don't want them installed with ``conan config install``

memsharded (approved):

memsharded (approved):

PR #4, opened 2023-03-01
Better documentation and CI support

uilianries:
- Update README with small example of conan custom command example. It could be moved to its folder, but the README is still small, so I don't think it worth moving it now.
- Added testing section and license on README
- Added tox support to automate pytest execution (ensure Python 3.6 support)
- Add Github Workflows to run test for each new PR
- CI Result: https://github.com/uilianries/conan-extensions/actions/runs/4305184791/jobs/7507266162


tox.ini
```diff
@@ -0,0 +1,24 @@
+[tox]
```
memsharded, line 1: Is this really necessary? We are not using tox in other Conan repos, I am fine if Python 3.6 is hardcoded in CI, and I'd prefer to keep it simpler

README.md
```diff
@@ -1,13 +1,42 @@
...
+conan migrate:convert-txt conanfile.txt > conanfile.py
+```
+
+### Testing
+
+To validate a new extension, it's possible to write a test which exercises its usage.
+You can use [pytest](https://docs.pytest.org) with [tox](https://tox.wiki/en/latest/) for testing and manegable virtual environment, producing a stable environment.
+
+```
+TOXENV=conan-latest tox
```
memsharded, line 37: Prefer ``pytest .`` testing

README.md
```diff
@@ -1,13 +1,42 @@
...
+# Conan Extensions
+This repository contains extensions for Conan, such as [custom commands](https://docs.conan.io/2/reference/extensions/custom_commands.html)
+and [deployers](https://docs.conan.io/2/reference/extensions/deployers.html),
 useful for different purposes, like artifactory tasks, conan-center-index, etc.
 
 The contents of this repository are *not* production ready, they are intended as an aiding guide,
 and you'll probably want to create your own custom commands and/or deployers taking these as a base to ensure they meet your needs.
 
-To install all the available extensions, run `conan config install https://github.com/conan-io/conan-extensions.git`.
+### Conan config as installer
```
memsharded, line 9:
```suggestion
### Conan config install
```

tox.ini
uilianries: not really needed, only in case need to validate multiple versions. Fine, python 3.6 is good enough, let me revert it

uilianries: Done, tox has been removed.

README.md
uilianries: Done, replaced by pytest usage.

README.md
uilianries: Done.

README.md
```diff
@@ -1,13 +1,44 @@
...
+conan config install https://github.com/conan-io/conan-extensions.git
+```
 
 Afterwards, running `conan --help` should show you all the custom commands available.
 
-The current commands are:
- - `migrate:convert-txt`: Gets a `conanfile.txt` as input and returns its equivalent `conanfile.py`
+### Custom Commands
+
+These are the current supported custom commands:
```
RubenRBS, line 21:
```suggestion
These are the currently supported custom commands:
```

RubenRBS (approved):
I like this! (Also, good to see that tox is not used now that we dropped it for 2.0)

PR #6, opened 2023-03-06
Add export_all command, useful in CCI

RubenRBS:
This PR adds `export_all_sources`, which has been pretty useful in CCI


extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,131 @@
...
+import textwrap
+import yaml
+
+from conan.api.output import ConanOutput
+from conan.cli.command import conan_command, OnceArgument
+from conan.errors import ConanException
+
+
+def output_json(result):
+    print(json.dumps({
```
memsharded, line 12: print probably shouldn't be used, but ``cli_out_write()``

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,131 @@
...
+
+        recipe_folder = os.path.join("recipes", recipe_name)
+        if not os.path.isdir(recipe_folder):
+            raise ConanException(f"Invalid user input: '{recipe_name}' folder does not exist")
+
+        config_file = os.path.join(recipe_folder, "config.yml")
+        if not os.path.isfile(config_file):
+            raise ConanException(f"The file {config_file} does not exist")
+
+        with open(config_file, "r") as file:
```
memsharded, line 90: I'd try to avoid extra nesting with:
```python
config = yaml.safe_load(open(config_file, "r").read())
```
or the like. But having all the processing of the contents inside the ``open`` context seems a bit too much.

extensions/hooks/hook_reduce_conandata.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def load_yml(conanfile, path):
+    if os.path.isfile(path):
+        return yaml.safe_load(load(conanfile, path))
+    return None
+
+
+def post_export(conanfile):
+    if conanfile.conf.get('user.hooks:reduce_conandata', check_type=bool, default=False):
```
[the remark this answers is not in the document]
memsharded: To discuss. Automatic activation of the hooks is not great for this repo, and I don't love the opt-in by a conf, seems too much.

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+    group.add_argument('-l', '--list', action=OnceArgument, help="YAML file with list of recipes to export")
+    group.add_argument('-p', '--path', action=OnceArgument, help="Path of the global recipe folder")
+    args = parser.parse_args(*args)
+
+    if args.name is not None:
+        recipes_to_export = [args.name]
+    elif args.list is not None:
+        recipes_to_export = yaml.safe_load(open(args.list, 'r'))["recipes"]
+    else:
+        recipes_to_export = os.listdir(args.path)
```
prince-chrismc, line 67: When would export all the recipes make sense from a user point of view?

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+    print("</table>")
+
+
+@conan_command(group="Conan Center Index", formatters={"text": lambda result: None, "json": output_json, "md": output_markdown})
+def export_all_versions(conan_api, parser, *args):
+    """
+    Export all version of either a single recipe, a list of recipes, or a global recipes folder
+    """
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('-n', '--name', action=OnceArgument, help="Name of the recipe to export")
```
prince-chrismc, line 57:
```suggestion
    group.add_argument('-n', '--name', action=OnceArgument, help="Name of the recipe to export. This will export only one recipe.")
```

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+    group.add_argument('-l', '--list', action=OnceArgument, help="YAML file with list of recipes to export")
```
prince-chrismc, line 58:
```suggestion
    group.add_argument('-l', '--list', action=OnceArgument, help="YAML file with list of recipes to export. All the recipes on the list will be exported")
```

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+    group.add_argument('-p', '--path', action=OnceArgument, help="Path of the global recipe folder")
```
prince-chrismc, line 59:
```suggestion
    group.add_argument('-p', '--path', action=OnceArgument, help="Path of the global recipe folder. This will export all recipes")
```

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+                if recipe_name not in exported:
+                    exported[recipe_name] = []
+                exported[recipe_name].append(ref)
+                exported_refs.append(ref)
+
+                exported_with_revision.append(f"{ref[0].name}/{ref[0].version}#{ref[0].revision}")
+            except Exception as e:
+                failed.update({f"{recipe_name}/{recipe_subfolder}": str(e)})
+
+    out.title("EXPORTED RECIPES")
```
prince-chrismc, line 113: The client does these in PascalCase and it would be nice to be consistent 🙏

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    out.title("FAILED TO EXPORT")
+    for key, value in failed.items():
+        out.info(f"{key}, {value}")
+
+    out.title("RECIPE REFERENCES")
+    for item in exported_with_revision:
+        out.info(f"{item}")
+
+    out.title("REFERENCE LIST")
```
prince-chrismc, line 125:
```suggestion
    out.title("VERSION LIST")
```

extensions/commands/cci/cmd_export_all_versions.py
```diff
@@ -0,0 +1,130 @@
...
+
+    versions_list = [f"{item[1]}" for item in exported_refs]
+    out.info(f'{versions_list}')
```
prince-chrismc, line 128: Maybe add some comments?

it's confusing why you'd print `name/version#revision` in the section above and then again dump the list of versions (with no recipe name)

I dont know why this is helpful in the CLI stderr but no included in the JSON output 🤔

extensions/commands/cci/cmd_list_v2_ready.py
```diff
@@ -0,0 +1,114 @@
...
+from conan.cli.command import conan_command, OnceArgument
+from conan.errors import ConanException
+
+
+def output_json(results):
+    print(json.dumps(results, indent=2))
+
+
+@conan_command(group="Conan Center Index", formatters={"json": output_json})
+def list_v2_ready(conan_api: conan.api.conan_api.ConanAPI, parser, *args):
```
prince-chrismc, line 17: I understand for us this is useful but how applicable is this to a wider audience and what happened when we required v2 support?

This wont help find all the old stale versions which users will see but not be able to use 🤔

prince-chrismc (commented):
Some feedback from the export command and big picture question for the list v2 :)

prince-chrismc:
Merge Conflicts :(

extensions/commands/cci/cmd_export_all_versions.py
RubenRBS: This was needed by @jcar87 to generate a list to then pass to tapaholes, but I admit that outside of that, that output might not be the most useful one. I've moved the output to its own text formatter for now

extensions/commands/cci/cmd_list_v2_ready.py
RubenRBS: Good last point, we can always iterate over the commands. This was needed in c3i so it got  split into its own PR (And merged without tests, we were in a hurry 😢) The question of how this command can bring value to the users is a good one, and we should think about it, definitely :)

prince-chrismc (approved):
i think this is a solid first pass :)

PR #7, opened 2023-03-06
Add extra commands coming from the docu and examples

RubenRBS:
Adds `cmd_clean` which comes from the "custom command" example in the docs

Missing tests, will un-draft after adding them


extensions/commands/management/cmd_clean.py
```diff
@@ -0,0 +1,53 @@
...
+from conan.cli.command import OnceArgument, conan_command
+from conans.client.userio import UserInput
+
+
+recipe_color = Color.BRIGHT_BLUE
+removed_color = Color.BRIGHT_YELLOW
+
+
+@conan_command(group="Custom commands")
+def clean(conan_api: ConanAPI, parser, *args):
```
memsharded, line 12: This needs a more specific name, there will be many commands.
Something like ``clean_old_revisions`` or the like

extensions/deployers/licenses_deploy.py
```diff
@@ -0,0 +1,11 @@
+from conan.tools.files import copy
+import os
+
+
+def deploy(graph, output_folder):
```
memsharded, line 5: Good as a start. In the future, I'd make it more complete, being able to report the licenses from teh recipe attributes, and report over skipped binaries (not necessary part of the current product)

extensions/commands/management/cmd_clean.py
```diff
@@ -0,0 +1,53 @@
+from conan.api.conan_api import ConanAPI
```
memsharded, line 1: The name of the "layer" that is "management" probably a bit too generic, but I guess that until we have more commands it is difficult to have a better naming and structure.

memsharded (approved):
LGTM, only small concern is naming.

extensions/deployers/licenses_deploy.py
```diff
@@ -0,0 +1,11 @@
...
+    """Copies all the licenses from all the dependencies of the recipe"""
+    for name, dep in graph.root.conanfile.dependencies.items():
+        copy(graph.root.conanfile, "(COPY|COPYING|LICENSE).*",
```
uilianries, line 8: I remember other exotic names, like the license name itself and a typo, french file named as `LISENSE`. I suggest you running `find recipes -type f "conanfile.py -exec grep licenses {} +` to have more idea about license file names that we have in CCI

README.md
```diff
@@ -28,6 +33,17 @@ Gets a `conanfile.txt` as input and outputs its equivalent `conanfile.py`
 conan migrate:convert-txt conanfile.txt > conanfile.py
 ```
 
+
+### Deployers
+
+#### [Deploy sources](extensions/deployers/sources_deploy.py)
```
uilianries, line 39: I would suggest a two line command example, one line the custom command and the second a `ls` without the output, only to indicate where they were installed.

README.md
```diff
@@ -20,6 +20,11 @@ Afterwards, running `conan --help` should show you all the custom commands avail
 
 These are the currently supported custom commands:
 
+#### [Clean](extensions/commands/management/cmd_clean.py)
+
+Deletes (from the local cache or remotes) every recipe and package revision,
```
uilianries, line 25: I would suggest adding the command line as example here.

extensions/deployers/licenses_deploy.py
RubenRBS: I left the deployers for another PR to focus on the command

extensions/commands/management/cmd_clean.py
RubenRBS: Yes, I've been trying to find a better name but I'm not being able to find the right word either :/

extensions/deployers/licenses_deploy.py
RubenRBS: Will make sure to get as many as possible once I add them in the new PR

extensions/commands/maintenance/cmd_clean_old_revisions.py
```diff
@@ -0,0 +1,53 @@
...
+@conan_command(group="Extensions")
+def clean_old_revisions(conan_api: ConanAPI, parser, *args):
+    """
+    Deletes (from local cache or remotes) all recipe and package revisions but
+    the latest package revision from the latest recipe revision.
+    """
+    parser.add_argument('-r', '--remote', action=OnceArgument,
+                        help='Will remove from the specified remote')
+    parser.add_argument('--force', default=False, action='store_true',
+                        help='Remove without requesting a confirmation')
```
prince-chrismc, line 20: Confirming should be `-c` like `conan remove`

```
  -c, --confirm         Remove without requesting a confirmation
```

```suggestion
    parser.add_argument('-c', '--confirm', default=False, action='store_true',
                        help='Remove without requesting a confirmation')
```

RubenRBS:
I don't think this is needed anymore after https://github.com/conan-io/conan/pull/13505/files

PR #8, opened 2023-03-13
List v2 ready

RubenRBS:
Moves this command to its own pr from #6 so that it can be used in CCI sooner.
Added a `skip-binaries` flag to not get any information about binaries.

This still needs to hook enabled, but we still need to discuss whether it makes sense to add to the repo and enable them via confs, and as CCI already has the hook, let's open this PR without the hook and maybe add it in the future


uilianries (approved):
LGTM

memsharded (approved):

PR #9, opened 2023-03-13
[conan-v2-ready] Do not raise ConanException for missing config.yml

uilianries:
- Add command line example on README
- ConanCenterIndex has projects missing `config.yml` which cause break exit when using the custom commands. Better ignore them and mark as error message instead.
- ~Add an argument for an output file, because on regular terminal is easy by using `>` to redirect an output to file, but to run this command in Jenkins as part of a commands wrapper, it will be a mess.~

Related to #8

### DIFF
--- a/extensions/commands/cci/cmd_list_v2_ready.py
+++ b/extensions/commands/cci/cmd_list_v2_ready.py
@@ -13,6 +13,11 @@ def output_json(results):
     print(json.dumps(results, indent=2))
 
 
+def output_file_json(file_path, results):
+    with open(file_path, 'w') as f:
+        json.dump(results, f)
+
+
 @conan_command(group="Conan Center Index", formatters={"json": output_json})
 def list_v2_ready(conan_api: conan.api.conan_api.ConanAPI, parser, *args):
     """
@@ -22,6 +27,7 @@ def list_v2_ready(conan_api: conan.api.conan_api.ConanAPI, parser, *args):
     parser.add_argument('path', help="Path to global recipes folder")
     parser.add_argument('-r', '--remote', action=OnceArgument, help="Remote to query against")
     parser.add_argument('-p', '--profiles', action='append', default=list(), help="List of profiles to run graph info with if --skip-binaries is false")
+    parser.add_argument('-f', '--output-file', action=OnceArgument, help="A JSON file path to store the output")
     parser.add_argument('--skip-binaries', action='store_true', default=False, help="Skip binary checking for packages")
     args = parser.parse_args(*args)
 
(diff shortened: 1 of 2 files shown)


RubenRBS (approved):

extensions/commands/cci/cmd_list_v2_ready.py
```diff
@@ -13,6 +13,11 @@ def output_json(results):
     print(json.dumps(results, indent=2))
 
 
+def output_file_json(file_path, results):
```
memsharded, line 16: Command formatters shouldn't write to disk. Instead they should to print to stdout and file redirect to be used to create a file.

extensions/commands/cci/cmd_list_v2_ready.py
```diff
@@ -22,6 +27,7 @@ def list_v2_ready(conan_api: conan.api.conan_api.ConanAPI, parser, *args):
     parser.add_argument('path', help="Path to global recipes folder")
     parser.add_argument('-r', '--remote', action=OnceArgument, help="Remote to query against")
     parser.add_argument('-p', '--profiles', action='append', default=list(), help="List of profiles to run graph info with if --skip-binaries is false")
+    parser.add_argument('-f', '--output-file', action=OnceArgument, help="A JSON file path to store the output")
```
memsharded, line 30: We decided to use file redirection for this, not an extra argument, lets follow the Conan conventions.

memsharded (changes requested):

extensions/commands/cci/cmd_list_v2_ready.py
```diff
@@ -39,7 +45,8 @@ def list_v2_ready(conan_api: conan.api.conan_api.ConanAPI, parser, *args):
         recipe_folder = os.path.join("recipes", recipe_name)
         config_file = os.path.join(recipe_folder, "config.yml")
         if not os.path.exists(config_file):
-            raise ConanException(f"The file {config_file} does not exist")
+            out.error(f"The file {config_file} does not exist")
```
prince-chrismc, line 48: What recipes are missing the `config.yml` I would except all of them to have one 🤔

uilianries: it's not really needed because it was a way to say to jenkins which versions should be built and where they are located. On the beginning of CCI config.yml was more optional, because instead of `all` folder,  we used the version as folder name and Jenkins should be able to understand it.

uilianries:
@memsharded Done, reverted file output feature.

extensions/commands/cci/cmd_list_v2_ready.py
memsharded: I am not sure about this. If it will fail if the file is not there in C3I, then shouldn't this be a hard error? I am fine with either, as long as you all have it clear.

memsharded (approved):

uilianries:
Closing this one, as config.yml should be available on CCI.

PR #11, opened 2023-03-28
Artifactory commands layer POC

czoido:
This is just a very basic proof of concept around a command layer to:

- Create and upload BuildInfo to Artifactory. At the moment it's just a very basic implementation considering recipe and package artifacts only but not the dependencies.
- Modify artifacts properties in Artifactory

This is also adding two repositories for testing so that we can do real functional tests with two remotes.

This is also installing conan from the release/2.0 branch to circunvent a custom commands bug not yet release in 2.0.2.

```
> conan art:build-info --help
usage: conan build-info [-h] [-v [V]] {create,delete,get,promote,upload} ...

Manages JFROG BuildInfo

positional arguments:
  {append,create,delete,get,promote,upload}
                        sub-command help
    append              Append published build to the build info.
    create              Creates BuildInfo from a Conan graph json from a conan install or create.
    delete              Removes builds stored in Artifactory. Useful for cleaning up old build info data.
    get                 Get Build Info information.
    promote             Promote the BuildInfo from the source to the target repository.
    upload              Uploads BuildInfo json to repository.

> conan art:property --help  
usage: conan property [-h] [-v [V]] {add,set} ...

Sets artifacts properties in Artifactory.

positional arguments:
  {add,build-info-add,set}
                        sub-command help
    add                 Append properties for artifacts under a Conan reference recursively.
    build-info-add      Load a Build Info JSON and add the build.number and build.name properties to all the artifacts present in the JSON.
    set                 Set properties for artifacts under a Conan reference recursively.
```


danimtb (commented):
Looking good!

tests/test_convert_txt.py
```diff
@@ -2,12 +2,15 @@
 import textwrap
 import os
 
+from tools import load, save, run
```
prince-chrismc, line 5: Can this be moved to a separate PR 🙏 I need it for my new test code too ❤️

tests/test_artifactory_commands.py
```diff
@@ -0,0 +1,102 @@
...
+    build_number = "1"
+
+    run(f"conan config install {repo}")
+    run("conan new cmake_lib -d name=mypkg -d version=1.0 --force")
+
+    run("conan create . --format json -tf='' -s build_type=Release > create_release.json")
+    run("conan create . --format json -tf='' -s build_type=Debug > create_debug.json")
+
+    run("conan remove mypkg* -c -r extensions-stg")
+    run("conan remove mypkg* -c -r extensions-prod")
```
prince-chrismc, line 60: Maybe add somethnig to the conanfile so a new revision is made and then we dont have to clean up?

Maybe the cleanup can be the global setup at the top? Seems out of place to have it before anything is uplaoded

tests/test_artifactory_commands.py
```diff
@@ -0,0 +1,102 @@
...
+        os.chdir(cwd)
+        os.environ.clear()
+        os.environ.update(old_env)
+
+
+def test_build_info_create():
+    repo = os.path.join(os.path.dirname(__file__), "..")
+
+    build_name = "mybuildinfo"
+    build_number = "1"
```
prince-chrismc, line 51: Could the build number be tied to the GitHub Actions run number?

tests/test_artifactory_commands.py
```diff
@@ -0,0 +1,102 @@
...
+    run(f'conan art:property build-info-add {build_name}_aggregated.json {os.getenv("ART_URL")} extensions-stg --user="{os.getenv("CONAN_LOGIN_USERNAME_EXTENSIONS_STG")}" --password="{os.getenv("CONAN_PASSWORD_EXTENSIONS_STG")}"')
+
+    run(f'conan art:build-info get {build_name}_release {build_number} {os.getenv("ART_URL")} --user="{os.getenv("CONAN_LOGIN_USERNAME_EXTENSIONS_STG")}" --password="{os.getenv("CONAN_PASSWORD_EXTENSIONS_STG")}"')
+    run(f'conan art:build-info get {build_name}_debug {build_number} {os.getenv("ART_URL")} --user="{os.getenv("CONAN_LOGIN_USERNAME_EXTENSIONS_STG")}" --password="{os.getenv("CONAN_PASSWORD_EXTENSIONS_STG")}"')
+    run(f'conan art:build-info get {build_name}_aggregated {build_number} {os.getenv("ART_URL")} --user="{os.getenv("CONAN_LOGIN_USERNAME_EXTENSIONS_STG")}" --password="{os.getenv("CONAN_PASSWORD_EXTENSIONS_STG")}"')
+
+    run(f'conan art:build-info promote {build_name}_aggregated {build_number} {os.getenv("ART_URL")} extensions-stg extensions-prod --user="{os.getenv("CONAN_LOGIN_USERNAME_EXTENSIONS_STG")}" --password="{os.getenv("CONAN_PASSWORD_EXTENSIONS_STG")}"')
+
+    run('conan remove mypkg* -c')
```
prince-chrismc, line 86: why a local clean if the promotion is done server side?

czoido: The local clean is because later I'm going to do a conan install from prod repo and I want to make sure that the install succeeds because the package comes from the remote and not because it's already in the cache. II'll add some comments to make it clear.

tests/test_artifactory_commands.py
```diff
@@ -0,0 +1,102 @@
...
+
+    run("conan create . --format json -tf='' -s build_type=Release > create_release.json")
+    run("conan create . --format json -tf='' -s build_type=Debug > create_debug.json")
+
+    run("conan remove mypkg* -c -r extensions-stg")
+    run("conan remove mypkg* -c -r extensions-prod")
+
+    run("conan upload mypkg/1.0 -c -r extensions-stg")
+
+    run(f'conan art:build-info create create_release.json {build_name}_release {build_number} > {build_name}_release.json')
```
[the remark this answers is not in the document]
czoido: Very good observation, yes, it requires the upload because we need to have the artifacts that are uploaded to Artifactory in the local cache and that happens in the upload process. I had a [TODO](https://github.com/conan-io/conan-extensions/blob/d6623f3991d5f66b6e5cb8574e721d0b2198856f/extensions/commands/art/cmd_build_info.py#L101) to raise in case the upload had not been made before this, I'll add the code to do it and some comments.

tests/test_artifactory_commands.py
czoido: I think this is fine to avoid bloating the repo with packages from the tests, I may refactor this later in a fixture or something like that...

tests/test_artifactory_commands.py
czoido: I think it's better to use this invented values for these tests (that are just to test the basics) but I think it's a good idea to add some more real tests (maybe they should be in another repo testing complex CI flows with Conan using this set of commands)

tests/test_artifactory_commands.py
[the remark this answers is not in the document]
czoido: Added the `test_fail_if_not_uploaded` test

tests/test_artifactory_commands.py
czoido: Added some comments.

danimtb (approved):

extensions/commands/art/cmd_build_info.py
```diff
@@ -0,0 +1,359 @@
...
+            md5.update(data)
+            sha1.update(data)
+            sha256.update(data)
+    return md5.hexdigest(), sha1.hexdigest(), sha256.hexdigest()
+
+
+def get_artifacts(cache_folder, remote_path):
+    artifacts = []
+    dl_folder = Path(cache_folder).parents[0] / "d"
+    # TODO: throw error if d is empty? Force that the artifacts where uploaded before getting the BuildInfo?
```
RubenRBS, line 101:
```suggestion
    # TODO: throw error if d is empty? Force that the artifacts were uploaded before getting the BuildInfo?
```

RubenRBS (approved):
👍

extensions/commands/art/cmd_property.py
```diff
@@ -0,0 +1,221 @@
...
+        recipe_ref = RecipeReference.loads(ref)
+        package_ref = None
+
+    rrev_path = f"/_/{recipe_ref.revision}" if recipe_ref.revision else ""
+    pkgid_path = f"/package/{package_ref.package_id}" if package_ref and package_ref.package_id else ""
+    prev_path = f"/{package_ref.revision}" if package_ref and package_ref.revision else ""
+    return f"_/{recipe_ref.name}/{recipe_ref.version}{rrev_path}{pkgid_path}{prev_path}"
+
+
+def api_request(type, request_url, user=None, password=None, apikey=None, json_data=None):
```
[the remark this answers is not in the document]
czoido: Yes, thanks for pointing that out, I mentioned that here: https://github.com/conan-io/conan-extensions/issues/19 we should adress this in the client.

PR #15, opened 2023-03-29
Runtime deployer

prince-chrismc:


memsharded (commented):
There are tests in this repo, a small test that covers it would be good.

README.md
```diff
@@ -30,6 +30,14 @@ Commands useful for migrations
 Commands useful in Conan Center Index or its forks
 
 
+### Deployer
+
+These are the currently supported custom commands:
```
RubenRBS, line 35:
```suggestion
These are the currently supported custom deployers:
```

But we're now creating a README per folder, so one in extensions/deploy would be nice. The file won't be copied to the user's cache

prince-chrismc:
> There are tests in this repo, a small test that covers it would be good.

I can not figure out why the installed deployer is not being found, I've stole the test from https://github.com/conan-io/conan/blob/dc17e1c113ef2773a23a25e317bd3d218aa02a9a/conans/test/functional/command/test_install_deploy.py#L70 🤔

prince-chrismc:
Screenshot_20230410_213203_Chrome

Theres a freaking space.

czoido:
> @czoido The artifactory/build info test from #11 is failing with missing login 😕 May you please take a peek?
> 

Hi @prince-chrismc! I think that's happening because the secrets to access to Artifactory are only available in the main repository but not on the forks. That's something we should handle, maybe in the meantime you can push the branch to the main repo and then delete the branch after the PR is merged?

prince-chrismc:
This will block all PRs from contributors which there's already 2 of :)

prince-chrismc:
https://github.com/conan-io/conan-extensions/pull/25 Passed 👍 that was definitely it

PR #24, opened 2023-04-05
Add BuildInfo dependencies support

czoido:
Still a lot to refactor and clean, it's still a bit messy.

Found issues:

- Looks like BuildInfo promotion with dependencies does not work, it works ok for all artifacts declared in the modules as dependencies except for the case of the `conaninfo.txt` this happens because it appears two times as a dependency with the same hash, because the package id is the same for two different revisions. Looks like as the hash is the same and it appears multiple times Artifactory skips the copy the second time.


extensions/commands/art/README.md
```diff
@@ -0,0 +1,101 @@
...
+    promote             Promote the BuildInfo from the source to the target repository.
+    upload              Uploads BuildInfo json to repository.
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -v [V]                Level of detail of the output. Valid options from less verbose to more verbose: -vquiet, -verror, -vwarning, -vnotice, -vstatus, -v or -vverbose, -vv or -vdebug, -vvv or
+                        -vtrace
+```
+
+#### [conan art:build-info](cmd_property.py)
```
czoido, line 36:
```suggestion
#### [conan art:property](cmd_property.py)
```

extensions/commands/art/cmd_build_info.py
```diff
@@ -95,85 +94,276 @@ def get_hashes(file_path):
...
+        Function to get artifact information, those artifacts can be added as artifacts of a
+        module or as artifacts from dependencies and depending on that the format is
+        different. For artifacts of modules they have the keys 'name' and 'path'. If they come
+        from a dependency they have an 'id' instead of 'name' and they don't have 'path'.
+        """
+
+        assert artifact_type in ["recipe", "package"]
+
+        if artifact_type == "recipe":
+            artifacts_names = ["conan_sources.tgz", "conanfile.py", "conanmanifest.txt"]
```
memsharded, line 204: might be missing the export.tgz

memsharded:
> Looks like BuildInfo promotion with dependencies does not work, it works ok for all artifacts declared in the modules as dependencies except for the case of the conaninfo.txt this happens because it appears two times as a dependency with the same hash, because the package id is the same for two different revisions. Looks like as the hash is the same and it appears multiple times Artifactory skips the copy the second time.

So the new ``path`` field is not able to dissambiguate duplicated hashes?

danimtb (approved):

PR #28, opened 2023-04-21
Add `cci:promote-graph`

RubenRBS:
This comes from work done to debug the gstream issue with @jwillikers 's gstreamer recipe. For that, I needed to move the recipe and its dependencies to another repo to be able to isolate and debug much faster, but missing dependencies were an issue. In the past we have done this by hand, but figured I'd save time for the next time this operation is needed:

### DIFF
--- a/extensions/commands/art/cmd_promote_graph.py
+++ b/extensions/commands/art/cmd_promote_graph.py
@@ -1,0 +1,111 @@
+import json
+import os.path
+
+from conan.api.conan_api import ConanAPI
+from conan.api.output import ConanOutput
+from conan.cli.command import conan_command
+from conans.model.recipe_ref import RecipeReference
+from conans.model.package_ref import PkgReference
+from conan.api.model import MultiPackagesList
+from conan.errors import ConanException
+
+from utils import api_request, assert_server_or_url_user_password
+from cmd_server import get_url_user_password
+
+
+def _get_path_from_rrev(rrev):
+    recipe_ref = RecipeReference.loads(rrev)
+    rrev_path = f"/_/{recipe_ref.revision}" if recipe_ref.revision else ""
+    return f"_/{recipe_ref.name}/{recipe_ref.version}{rrev_path}"
+
+
+def _get_path_from_ref(ref):
+    try:
+        package_ref = PkgReference.loads(ref)
+        recipe_ref = package_ref.ref
+    except ConanException:
+        recipe_ref = RecipeReference.loads(ref)
+        package_ref = None
+
+    rrev_path = f"/_/{recipe_ref.revision}" if recipe_ref.revision else ""
+    pkgid_path = f"/package/{package_ref.package_id}" if package_ref and package_ref.package_id else ""
+    prev_path = f"/{package_ref.revision}" if package_ref and package_ref.revision else ""
+    return f"_/{recipe_ref.name}/{recipe_ref.version}{rrev_path}{pkgid_path}{prev_path}"
+
+
+def _request(url, user, password, request_type, request_url):
+    try:
+        return json.loads(api_request(request_type, f"{url}{request_url}", user, password))
+    except Exception as e:
+        raise ConanException(f"Error requesting {request_url}: {e}")
+
+
+def _promote_path(url, user, password, origin, destination, path):
+    ConanOutput().subtitle(f"Promoting {path}")
+    # The copy api creates a subfolder if the destination already exists, need to check beforehand to avoid this
+    try:
+        # This first request will raise a 404 if no file is found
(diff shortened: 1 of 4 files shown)


RubenRBS:
Still working on your PR timeout issue @jwillikers :)

jwillikers:
> Still working on your PR timeout issue @jwillikers :)

Thanks! And actually, I'm really excited to see you adding this feature because it should be really valuable where I work!

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,111 @@
...
+from conan.api.model import MultiPackagesList
+from conan.errors import ConanException
+
+from utils import api_request, assert_server_or_url_user_password
+from cmd_server import get_url_user_password
+
+
+def _get_path_from_rrev(rrev):
+    recipe_ref = RecipeReference.loads(rrev)
+    rrev_path = f"/_/{recipe_ref.revision}" if recipe_ref.revision else ""
```
memsharded, line 18: Is this hardcoding packages without user/channel? It should work with private packages with user/channel too.

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,111 @@
...
+    return f"_/{recipe_ref.name}/{recipe_ref.version}{rrev_path}"
+
+
+def _get_path_from_ref(ref):
```
memsharded, line 22: Functions to map to Artifactory paths most like can be common to other commands too?

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,111 @@
...
+    # Only artifactory pro edition supports this feature
+    response = _request(url, user, password, "get", "api/system/version")
+    if response["license"] == "Artifactory Community Edition for C/C++":
+        raise ConanException("Direct graph promotion is only supported in Artifactory Pro. As an alternative, use conan download + conan upload with the pkglist feature")
+
+    listfile = os.path.realpath(args.list)
+    multi_package_list = MultiPackagesList.load(listfile)
+
+    remotes = list(multi_package_list.lists.keys())
+    if len(remotes) > 1:
```
memsharded, line 84: I think it shouldn't be a problem to have a package list with more than 1 remote. Use case, having a package lists created with multiple remotes, and we want to promote all of them from different repos to another "production" repo.

memsharded: So I would use the ``--remotes`` inputs as other commands.

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,111 @@
...
+@conan_command(group="Artifactory")
+def promote_graph(conan_api: ConanAPI, parser, *args):
+    """
+    Promote a pkglist file from an origin repository to a destination repository, without downloading the packages locally
+    """
+    parser.add_argument("origin", help="Artifactory origin repository.")
+    parser.add_argument("destination", help="Artifactory destination repository.")
+    parser.add_argument("list", help="Package list file to promote")
+
+    parser.add_argument("--server", help="Server name of the Artifactory to get the build info from")
```
memsharded, line 64: No longer using build info? can be removed?

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,111 @@
...
+        ConanOutput().success("Promoted file")
+
+
+@conan_command(group="Artifactory")
+def promote_graph(conan_api: ConanAPI, parser, *args):
+    """
+    Promote a pkglist file from an origin repository to a destination repository, without downloading the packages locally
+    """
+    parser.add_argument("origin", help="Artifactory origin repository.")
+    parser.add_argument("destination", help="Artifactory destination repository.")
```
memsharded, line 61: Maybe we want to process the destination from a Conan remote (from name, get the URL, and deduce the target from from it)?

memsharded (commented):
Looks good, simple and I think it can be useful

czoido:
A discussion from the cppland Slack channel:

Consider the case that the production repo you are promoting to does not have overwrite permissions. In the case of Build promotion via the extension, as it's using the Artifactory API https://jfrog.com/help/r/jfrog-rest-apis/build-promotion that does not consider this it will fail. We may consider adding some option to avoid overwriting on promotions in case the user requests it.

markferry:
> Consider the case that the production repo you are promoting to does not have overwrite permissions.

For context, this arose from `art:build-info promote` in which the buildinfo referenced exports (sources.tgz etc) which were already in the target repo - so could not be overwritten.

PR #34, opened 2023-05-03
Simplify art:build-info UX

czoido:
It was redundant to set the properties after or before uploading the BuildInfo, you can do that in just one step.
Maybe we could remove the build-info add command or it could serve to add arbitrary properties to artifacts present in a BuildInfo


danimtb (approved):

danimtb:
I think it is fine to keep the command to add properties although it is not strictly related to build info. I think the command could be useful for storing metadata in properties

prince-chrismc:
👍  For saving the command it for now, I think there's a good use cases for having user specific properties they need to track?

czoido:
> 👍 For saving the command it for now, I think there's a good use cases for having user specific properties they need to track?

Yes, I think that could be useful, I'll add an argument for that.

Issue #17, opened 2023-04-03
Do conan art:build-info append from builds published in Artifactory


czoido:
Currently is made merging the json files but it would make more sense to use the build name and number to download the builds from Artifactory so we make sure that they are already published.

PR #35, opened 2023-05-04
Do the append of BuildInfos from already published BuildInfos instead from the JSON files

czoido:
Closes: https://github.com/conan-io/conan-extensions/issues/17

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -592,21 +592,34 @@ def build_info_append(conan_api: ConanAPI, parser, subparser, *args):
     subparser.add_argument("build_name", help="The current build name.")
     subparser.add_argument("build_number", help="The current build number.")
 
-    subparser.add_argument("--build-info", help="JSON file for the Build Info. You can add multiple files " \
-                                                "like --build-info=release.json --build-info=debug.json",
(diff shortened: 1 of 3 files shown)


danimtb (approved):
Looking good! This will greatly improve the UX 😄

extensions/commands/art/cmd_build_info.py
```diff
@@ -592,21 +592,34 @@ def build_info_append(conan_api: ConanAPI, parser, subparser, *args):
     subparser.add_argument("build_name", help="The current build name.")
     subparser.add_argument("build_number", help="The current build number.")
 
-    subparser.add_argument("--build-info", help="JSON file for the Build Info. You can add multiple files " \
-                                                "like --build-info=release.json --build-info=debug.json",
+    subparser.add_argument("url", help="Artifactory url, like: https://<address>/artifactory")
+
+    subparser.add_argument("--user", help="user name for the repository")
+    subparser.add_argument("--password", help="password for the user name")
+    subparser.add_argument("--apikey", help="apikey for the repository")
```
prince-chrismc, line 599: Can custom commands read from the `source_credentials.json` file?

czoido: Not yet, but credentials management in this command I think that is definitely something we have to improve because it's tricky to add all those arguments. I'll create an issue so that we can think how to improve this.

PR #37, opened 2023-05-08
Fix wrong name for art:property command

WilliamBehrens:
The readme listed `art:build-info` twice when the second line should have been `art:property`.


RubenRBS (approved):
Good catch :)

czoido (approved):

PR #39, opened 2023-05-08
art:build-info Quick test for uploading to project specific build info

### LINKED ISSUE #38: Add projects support to art:build-info commands
Trying to organize and prep a demo, I wants to use a Project in Artifactory to separate my artifacts and builds.

image

How the URLs are missing the query parm for this https://jfrog.com/help/r/jfrog-rest-apis/build-promotion `GET /api/build?project='projectKey'`

prince-chrismc:
closes #38

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -8,7 +8,7 @@ from pathlib import Path
 import requests
 
 from conan.api.conan_api import ConanAPI
-from conan.api.output import cli_out_write
+from conan.api.output import cli_out_write, ConanOutput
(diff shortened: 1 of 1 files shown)


prince-chrismc:
This feature allows user to have a per-project view of the build and artifacts they are working with.
image

Which hides the previous attempts

image

czoido (commented):
Great!
I think we should also add the project argument to the  `art:build-info append` and `art:build-info add` commands? Also, could we add some tests?

prince-chrismc:
Yes! ❤️ I got distracted with other priorities so lets plan this for next sprint? I do not think it's blocker

PR #43, opened 2023-05-18
[feature] Implement --remote flag in art:build-info commands

danimtb:
On top of https://github.com/conan-io/conan-extensions/pull/42 (that should be merged first)

If you want to review this, check only the `extensions/commands/art/cmd_build_info.py` file diff.

- Adds `--remote` flags to the commands.
- Added helper function to assert for proper flags
- Added helper to get remote information
- Removed api key flag and request implementation as it is deprecated: [JFrog API Key Deprecation Process](https://jfrog.com/help/r/jfrog-platform-administration-documentation/api-key)

Issue #36, opened 2023-05-05
Improve credentials management in conan art:build-info commands


czoido:
I think we should provide a more convinient way to manage credentials than passing always as parameters. Maybe reading from [source_credentials.json](https://docs.conan.io/2/reference/config_files/source_credentials.html) or trying to read environment variables with some known format?

danimtb:
I have been taking a look at JFrog CLI to see how it handles it.

Command help:
```
$ jf config add -h

Name:
  jf c add - Adds a server configuration.

Usage:
  jf config add
  jf config add <server ID>

Options:
  --access-token            [Optional] JFrog Platform access token.
  --access-token-stdin      [Default: false] Set to true if you'd like to provide the access token via stdin.
  --artifactory-url         [Optional] JFrog Artifactory URL.
  --basic-auth-only         [Default: false] Set to true to disable replacing username and password/API key with automatically created access token that's refreshed hourly. Username and password/API key will still be used with commands which use external tools or the JFrog Distribution service. Can only be passed along with username and password/API key options.
  --client-cert-key-path    [Optional] Private key file for the client certificate in PEM format.
  --client-cert-path        [Optional] Client certificate file in PEM format.
  --distribution-url        [Optional] JFrog Distribution URL.
  --enc-password            [Default: true] If set to false then the configured password will not be encrypted using Artifactory's encryption API.
  --insecure-tls            [Default: false] Set to true to skip TLS certificates verification, while encrypting the Artifactory password during the config process.
  --interactive             [Default: true, unless $CI is true] Set to false if you do not want the config command to be interactive. If true, the --url option becomes optional.
  --mission-control-url     [Optional] JFrog Mission Control URL.
  --overwrite               [Default: false] Overwrites the instance configuration if an instance with the same ID already exists.
  --password                [Optional] JFrog Platform password or API key.
  --password-stdin          [Default: false] Set to true if you'd like to provide the password via stdin.
  --pipelines-url           [Optional] JFrog Pipelines URL.
  --ssh-key-path            [Optional] SSH key file path.
  --ssh-passphrase          [Optional] SSH key passphrase.
  --url                     [Optional] JFrog platform URL.
  --user                    [Optional] JFrog Platform username.
  --xray-url                [Optional] JFrog Xray URL.
```

Command example:
```
$ jf config add --artifactory-url https://c3i.jfrog.io/artifactory --user myuser --password mypassword
Choose a server ID: 1
[...]

PR #45, opened 2023-05-22
[feature] Implement --server flag in art:build-info & art:property commands

danimtb:
closes #43
closes https://github.com/conan-io/conan-extensions/issues/36

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -15,8 +15,6 @@ from conan.errors import ConanException
 from conans.model.recipe_ref import RecipeReference
 from conan import conan_version
 
-from cmd_server import read_servers
-
 SERVERS_FILENAME = ".art-servers"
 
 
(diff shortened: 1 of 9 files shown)


extensions/commands/art/cmd_build_info.py
```diff
@@ -14,6 +15,8 @@
 from conans.model.recipe_ref import RecipeReference
 from conan import conan_version
 
+SERVERS_FILENAME = ".art-servers"
```
franramirez688, line 20: I'd try to use another name like `SERVERS_FILE_EXTENSION` or ` SERVERS_FILE_EXT` to simplify...

Furthermore, I saw you're getting the path to this `servers` file, so I think this could go as a global variable:

```python
SERVERS_FILE_PATH = os.path.join(os.path.dirname(__file__), ".art-servers")
```
(even dropping `SERVERS_FILE_EXTENSION` if it's only used here)

danimtb: it is not only the file extension but the name of the actual file (it is an encoded file)

RubenRBS (approved):
Looks nice to me, thanks for also updating the docs :)

czoido (approved):
Really nice!

PR #51, opened 2023-06-01
[ci] Run test with latest conan version and current develop version

danimtb:
Added additional steps to test with latest conan version as well

### DIFF
--- a/.github/workflows/test_conan_extensions.yml
+++ b/.github/workflows/test_conan_extensions.yml
@@ -9,6 +9,9 @@ jobs:
       CONAN_LOGIN_USERNAME_EXTENSIONS_STG: ${{ secrets.CONAN_USER }}
       CONAN_PASSWORD_EXTENSIONS_STG: ${{ secrets.CONAN_PASSWORD }}
       ART_URL: "https://conanv2beta.jfrog.io/artifactory"
+    strategy:
+      matrix:
+        conan-version: ['latest', 'develop']
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python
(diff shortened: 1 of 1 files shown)


.github/workflows/test_conan_extensions.yml
```diff
@@ -18,7 +21,14 @@ jobs:
...
-        pip install pytest git+https://github.com/conan-io/conan.git@release/2.0
-    - name: Conan latest - Python 3.6
+        pip install pytest
+    - name: Install Conan
+      run: |
+        if [ "${{ matrix.conan-version }}" == "latest" ]; then
+          pip install conan
+        elif [ "${{ matrix.conan-version }}" == "develop" ]; then
+          pip install git+https://github.com/conan-io/conan.git@release/2.0
+        fi
```
RubenRBS, line 31: What about something like this? Note that instead of develop we could use the branch name and have an easier time updating when we eventually move to release/2.1 

(Syntax not checked, sent from phone)

```suggestion
    - name: Install Latest Conan
      if: matrix.conan-version == "latest"
      run: pip install conan
   - name: Install ${{ matrix.conan-version }} 
     if: matrix.conan-version != "latest"
     run: pip install git+https://github.com/conan-io/conan.git@${{ matrix.conan-version }}
```

RubenRBS (approved):
Nice! A comment about how to prepare for bad eventual release/2.1 and such

.github/workflows/test_conan_extensions.yml
danimtb: this approach was not working as tasks were run in parallel and caused issues with artifactory

czoido (approved):

Issue #52, opened 2023-06-06
[feature request] Command to deploy macOS/iOS universal binaries


gmeeker:
See the long discussion here:
https://github.com/conan-io/conan/issues/1047

The simplest approach is to deploy the Conan dependencies, run lipo on everything, then use an Xcode project created by hand.  My example command is here:

https://github.com/gmeeker/conan-deploy-lipo

PR will follow shortly.

gmeeker:
I have a different attempt here
https://github.com/conan-io/conan-extensions/pull/197

I'll probably close the deploy based attempt.

PR #58, opened 2023-06-10
[feature] implement lipo deployer (#52)

gmeeker:
This is a different approach to #52 where the user deploys once per architecture. The previous attempt used a custom command that handled multiple deployments.

### DIFF
--- a/extensions/deployers/lipo/README.md
+++ b/extensions/deployers/lipo/README.md
@@ -1,0 +1,57 @@
+## Lipo deployers
+
+These deployers create universal binaries on macOS or iOS.
+They are similar to Conan's full_deploy deployer but run `lipo` to
+produce universal binaries. An Xcode project can use the `full_deploy`
+folder to compile a universal application.
+
+
+#### [lipo](lipo.py)
+
+This deployer is identical to `full_deploy` except single architecture universal
+binaries will be created and a subfolder will not be created with the arch name.
+
+
+#### [lipo_add](lipo_add.py)
+
+This deployer doesn't remove the existing `full_deploy` folder but adds an architecture
+to the existing universal binaries. It is an error to run this more than once for a given
+architecture. Universal binaries can only contain one binary per architecture. See
+`lipo -create` for more information.
+
+```sh
+$ conan install . --deploy=lipo . -s arch=x86_64
+$ conan install . --deploy=lipo_add . -s arch=armv8
+```
+
+Profiles can be used to handle additional settings, for example when the architectures
+have a different minimum deployment OS.
+
+
+## Sample profiles
+
+#### x86_64
+```
+[settings]
+arch=x86_64
+build_type=Release
+compiler=apple-clang
+compiler.cppstd=gnu17
+compiler.libcxx=libc++
+compiler.version=14
+os=Macos
+os.version=10.13
+```
+
+#### armv8
+```
+[settings]
+arch=armv8
+build_type=Release
+compiler=apple-clang
+compiler.cppstd=gnu17
+compiler.libcxx=libc++
+compiler.version=14
+os=Macos
+os.version=11.0
+```
--- a/extensions/deployers/lipo/lipo.py
+++ b/extensions/deployers/lipo/lipo.py
@@ -1,0 +1,31 @@
+import os
+import shutil
+
+from conan.tools.files import copy
+from conans.util.files import rmdir
+from conan.tools.apple import is_apple_os
+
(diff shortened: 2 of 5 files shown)


TheSin-:
I'm trying to test this out with conan 2.0.9.  I install them and see it installing

```
...
Copying file tools_lipo.py to /Users/thesin/.conan2/extensions/deployers/lipo
Copying file lipo.py to /Users/thesin/.conan2/extensions/deployers/lipo
Copying file lipo_add.py to /Users/thesin/.conan2/extensions/deployers/lipo
...
```

but I do not see an option for `install --deploy` in `install -h` and when I try to use it, I get an error about `--deploy`

```
conan install . --deploy=lipo . -s arch=armv8
...
conan install: error: ambiguous option: --deploy=lipo could match --deployer, --deployer-folder
```

I'm still trying to learn conan so sorry if this is a noob thing

TheSin-:
perfect thank you!  I'll continue testing now ;)

gmeeker:
It wasn't ambiguous in 2.0.6 but I've updated the README. If you have any feedback please let me know. I'm probably going to close the earlier custom command approach, but I've been waiting to see if anyone has thoughts about the two approaches.

WarGloom:
Can we create universal packages without local deploy? Honestly I don't like idea copying everything to build folder. 
I'm not sure I understand why you decide use deploy over hooks ?

WarGloom:
if run lipo deployer with generator CMakeToolchain, after lipo_add it overrides CMAKE_OSX_ARCHITECTURES  to last used arch

gmeeker:
The history in #59 is quite long now, and probably not worth reading in its entirety.
Most newcomers to Conan seem have the same initial thought: why can't I just set CMAKE_OSX_ARCHITECTURES and get universal binaries? Unfortunately there are thousands of recipes in CCI and many of them use other build systems (see OpenSSL). Some of these can be built by overriding CFLAGS with "-arch x86_64 -arch arm64" but the most reliable method is to simply build each architecture and run lipo on the resulting binaries.
I looked at different ways of automating this (without updating thousands of recipes) including hooks (not possible for all recipes) or modifying conan itself (very complicated and invasive).

This deployer is the simplest extension to start testing the lipo code and it still needs additional feedback. I haven't tested iOS at all. I hoped it would be merged by now, but maybe most people are just waiting for a full solution with a conanfile building a universal app?

Beyond this initial test as a deployer, I intend to investigate generating a universal CMake project as you mentioned above (which was possible in 1.x with toolchains but no longer seems to be officially supported). Then connecting the CMake generator to the universal dependencies.

I had a working 1.x solution here (for the universal dependency + CMake toolchain for final build) but it's very kludgy and uses Conan internals. I'd like to do something similar with the public API and new 2.0 features that is clean enough to be accepted to conan-extensions.
https://github.com/gmeeker/BuildConan

MartinDelille:
Hi @gmeeker !

I am very interested in testing your work as soon as my project will be conan 2 ready (my application is still using Rosetta).

WarGloom:
Ok, Great I've been able to build universal libraries and project 
except I have to use generator CMakeDeps
with CMakeToolchain it defines only the last arch used with lipo_add in cmake files

Greate work, than you, and please add to Readme how to proper install this extension. It's not obvious for newbie

garyo:
I need to build universal binaries for my conan dependencies on Mac; this looks like it would be useful. Looks it hasn't been merged into main yet; any update on that?

memsharded:
Hi all,

Sorry, this hasn't been reviewed yet, there are too many things ongoing and the backlog is just too large, we have been trying to prioritize things like the migration of ConanCenter to Conan 2.0.

I think we need to resume this, I'll try to bring this to the team attention hopefully in the next weeks. Thanks for your patience.

gmeeker:
It sounds like the deployer here is preferred over the custom command in #53 (the implementation details over argument parsing and loading conanfiles were never resolved cleanly.)

Also the iOS support here still needs to be tested. Support for xcarchive and xcframework might be necessary?

MartinDelille:
I'm now using conan 2.0 and I'm trying to make an univeral binary of my Qt MacOS application. If I understand well, this extension could help me, right ?

gmeeker:
Hmm, I actually started by looking at custom commands but there were technical issues that made it impossible in one go without a deploy. See here: https://github.com/conan-io/conan-extensions/issues/52

The lipo deployer here runs lipo multiple times, so your approach seems more consistent with the way Xcode produces universal binaries.

PR #66, opened 2023-08-03
Add new command sbom:cyclonedx

hedtke:
This replaces the proposal https://github.com/conan-io/conan/pull/14405

It closes https://github.com/conan-io/conan/issues/9787

It is a possible replacement for https://github.com/CycloneDX/cyclonedx-conan


memsharded:
If this could be a replacement for https://github.com/CycloneDX/cyclonedx-conan, it would be great to ping them and make them aware explicitly, maybe a ticket in their repo, to potentially centralize efforts somewhere. 

As the proposal is a new custom command, it is really not critical where is it, because it can be integrated easily with ``conan config install <url>``, so it could be either centralized in the CycloneDX org, or here in the Conan org, but as it seems they already put a great effort there, it makes sense to align with them.

memsharded:
Ok, I see that you already did in https://github.com/CycloneDX/cyclonedx-conan/issues/84. Commenting there.

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,73 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+
+#### [Create SBOM](cmd_create_sbom.py)
+
+For the graph JSON passed as parameter it creates an SBOM in CycloneDX 1.4 JSON format.
+
+**Parameters**
+- **graph_json**: Path to Conan generated graph JSON output file.
+
+```
+$ conan graph info -f json --requires=openssl/3.1.1 > openssl_graph.json
```
jkowalleck, line 48: > conan graph info -f json --requires=openssl/3.1.1 > openssl_graph.json

better use `cyclonedx:json` as a format.
there might be other json formats in the future.

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,86 @@
...
+
+def licenses(id):
+    """
+    see https://cyclonedx.org/docs/1.4/json/#components_items_licenses
+    """
+    if id is None:
+        return []
+    else:
+        return [{"license": {
+            "id": id
```
jkowalleck, line 17: per definition of `license` is not guaranteed to be a valid SPDX license ID.
read https://docs.conan.io/2/reference/conanfile/attributes.html#license

per CycloneDX JSON schema definition, this is the list of supported SPDX licenses: 
https://cyclonedx.org/schema/spdx.schema.json


so assuming every value was a SPDX licen id might lead to invalid CycloneDX documents.


i'd recommend doing the following: 
1. store a set of well-known licenses in the code
2. if the given license is in well-known licenses,
   then: return as `id`-license  as in https://cyclonedx.org/docs/1.4/json/#tab-pane_components_items_licenses_items_license_oneOf_i0
   else: return as `name`-license as in https://cyclonedx.org/docs/1.4/json/#tab-pane_components_items_licenses_items_license_oneOf_i1

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,86 @@
+import json
+from packageurl import PackageURL
```
jkowalleck, line 2: How is this dependency supposed to be installed? any information this is a requirement? 

If you use external dependencies in the first place, then you really should consider using https://pypi.org/project/cyclonedx-python-lib/ to generate the CycloneDX document, instead of crafting the document manually.

jkowalleck:
Hi, I am the maintainer of https://github.com/CycloneDX/cyclonedx-conan,
and a member of OWASP's CycloneDX Core Working Group

for the following reasons, i would invite you to chat and discuss in CycloneDX slack channel [#conan](https://cyclonedx.slack.com/archives/C02JF19QFUY)  
join via <https://cyclonedx.org/slack/invite>

I see some misconceptions and missing spots in this PR:
- missing `metadata.component` describing the thing the SBOM was built for
- dependency graph has no root, not connections, just nodes 
- optional component's namespace appears to be missing
- invalid component license values might render the whole JSON invalid

jkowalleck:
re: https://github.com/conan-io/conan-extensions/pull/66#issuecomment-1665396603
> [...]  it could be either centralized in the CycloneDX org, or here in the Conan org, but as it seems they already put a great effort there, it makes sense to align with them.

I will discuss with the CycloneDX core team a possible adoption of a Coann2 extension.
I will keep you updated here: https://github.com/CycloneDX/cyclonedx-conan/issues/84#issuecomment-1665400697

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,86 @@
...
+
+@conan_command(group="Recipe")
+def create_sbom(conan_api: ConanAPI, parser, *args):
+    """
+    creates an SBOM in CycloneDX 1.4 JSON format from a Conan graph JSON
+    """
+    parser.add_argument("graph_json", help="Path to Conan generated graph JSON output file.")
+    args = parser.parse_args(*args)
+    with open(args.graph_json, 'r') as f:
+        data = json.load(f)
```
andreas-hilti, line 72: You are loading the graph as a basic json document. I'm wondering whether it wouldn't be more sustainable to use the conan_api to load and access the data.

This is similar in sprit that you could maybe use https://pypi.org/project/cyclonedx-python-lib/ to write the SBOM (as pointed out by @jkowalleck ).

You can of course do everything manually, but this leads to duplication and higher maintenance costs.

Maybe you can also compare it to my draft:
https://github.com/andreas-hilti/cyclonedx-conan/tree/feat/conan_v2_draft

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,86 @@
...
+    qualifiers = {
+        "prev": node["prev"]
+    }
+    if node["user"]:
+        qualifiers["user"] = node["user"]
+    if node["channel"]:
+        qualifiers["channel"] = node["channel"]
+    if "#" in node["ref"]:
+        qualifiers["rref"] = node["ref"].split("#")[1]
+    qualifiers["repository_url"] = node["remote"] if node["remote"] else "https://center.conan.io"
```
andreas-hilti, line 34: Just a minor comment:
> If omitted, https://center.conan.io as default repository is assumed.

You might consider omitting the default value (as it is anyway assumed by the specification).

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    qualifiers = {
+        "prev": node["prev"]
+    }
+    if node["user"]:
+        qualifiers["user"] = node["user"]
+    if node["channel"]:
+        qualifiers["channel"] = node["channel"]
+    if "#" in node["ref"]:
+        qualifiers["rref"] = node["ref"].split("#")[1]
```
andreas-hilti, line 33: Note that the parsing of a reference is actually more complicated:
name/version@user/channel#rrev:pkgid#prev
https://docs.conan.io/2/reference/commands/list.html?highlight=prev
though I'm not sure whether they could appear in the graph json file.

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,86 @@
...
+        qualifiers=qualifiers)
+
+
+def component(n: dict) -> dict:
+    """
+    see https://cyclonedx.org/docs/1.4/json/#components
+    """
+    result = {
+        "type": "library",
+        "bom-ref": n["id"],
```
andreas-hilti, line 48: While the only formal requirement on the bom-ref is that it is unique within an SBOM, I'd personally prefer a more expressive bom-ref, e.g. the purl (which is also recommended to be used).
This might be very subjective, but at least it would help when manually inspecting the dependencies section.

andreas-hilti:
It would also be very helpful to add at least one or two tests.

extensions/commands/recipe/README.md
prince-chrismc: This is a standard command/feature of the client and it's no specific to the SBOM feature being added.

However I do like this suggestion for the line below, having a dedicated `--format` option or having a more specific name for the command 👍

extensions/commands/recipe/cmd_create_sbom.py
prince-chrismc: I think the most simple approach is to just always use `name` as a starting point, this has the most value for open-source packages.

It's unfortunate the schema is a hardcoded list that does not include the license expressions https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/ which we recommend currently https://github.com/conan-io/conan-center-index/commit/4c0c17a4a0956d2c3d8b8610ceebeb7746b5301e

extensions/commands/recipe/cmd_create_sbom.py
prince-chrismc: package id and prev are seperate feilds 

image

extensions/commands/recipe/cmd_create_sbom.py
andreas-hilti: Note that you can use license expressions, they are just not done under the "license" field, but under "expression":
https://cyclonedx.org/docs/1.4/json/#components_items_licenses_items_expression
e.g.
```
"licenses": [
  {
    "expression": "EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0"
  }
]
```
(https://github.com/CycloneDX/cyclonedx-dotnet-library/blob/main/tests/CycloneDX.Core.Tests/Json/v1.4/__snapshots__/SerializationTests.JsonRoundTripTest_valid-license-expression-1.4.json.snap)

extensions/commands/recipe/cmd_create_sbom.py
andreas-hilti: @prince-chrismc  Is the graph json format documented (other than just by this code https://github.com/conan-io/conan/blob/release/2.0/conans/client/graph/graph.py )?

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: this is true. the purl should be as short as possible.
tools and some OSS-Indexes had problems with long purls. so dropping standard/default values where you can helps shorten the purl.

extensions/commands/recipe/cmd_create_sbom.py
hedtke: I did not find in the docu how to specify requirements for conan extensions. How to do that? Can someone help here?

coderpatros:
> If this could be a replacement for https://github.com/CycloneDX/cyclonedx-conan, it would be great to ping them and make them aware explicitly, maybe a ticket in their repo, to potentially centralize efforts somewhere.
> 
> As the proposal is a new custom command, it is really not critical where is it, because it can be integrated easily with `conan config install <url>`, so it could be either centralized in the CycloneDX org, or here in the Conan org, but as it seems they already put a great effort there, it makes sense to align with them.

IMHO it would be better for this functionality to be maintained in the Conan ecosystem rather than the CycloneDX project. Package managers and build systems are complicated. And they all behave a bit differently. IMO it's much more likely to attract the required expertise and experience here. The CycloneDX team are happy to provide support, feedback, answer questions, etc.

extensions/commands/recipe/cmd_create_sbom.py
hedtke: I switched to the cyclonedx python package which takes care of that

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: @memsharded (how) can a conan2 extension add python dependencvies?

extensions/commands/recipe/cmd_create_sbom.py
hedtke: Changed: I set it now only when it is not the default value

extensions/commands/recipe/cmd_create_sbom.py
hedtke: Automatically fixed as I switched to the cyclonedx python package to create the BOM

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def licenses(id):
+    """
+    see https://cyclonedx.org/docs/1.4/json/#components_items_licenses
+    """
+    if id is None:
+        return None
+    else:
+        return [LicenseChoice(license=lFac.make_from_string(id))]
```
jkowalleck, line 31: perfect. this is exactly the correct usage 👍 
```py
lFac = LicenseFactory()
LicenseChoice(license=lFac.make_from_string(id))
```

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: > I switched to the cyclonedx python package which takes care of that

👍 
https://github.com/conan-io/conan-extensions/pull/66/files#r1286968212

extensions/commands/recipe/cmd_create_sbom.py
hedtke: I changed it: I use `common_graph_args` now, so the first step (create a graph JSON) is not longer required

hedtke:
> It would also be very helpful to add at least one or two tests.

Test added

hedtke:
Ready for next review round. I think I addressed everything. Thank you all for the helpful comments!

extensions/commands/recipe/cmd_create_sbom.py
prince-chrismc: We are very strict about adding dependencies, this should be avoid as much as possible https://github.com/conan-io/conan/pull/14405#discussion_r1280335011

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Creates a PURL following https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#conan
+    """
+    return PackageURL(
+        type="conan",
+        name=node.conanfile.name,
+        version=node.conanfile.version,
+        qualifiers={
+            "prev": node.prev,
+            "rref": node.ref.revision,
```
andreas-hilti, line 47: @hedtke  I think you need to check a couple of None values.
For instance the following conanfile.txt will lead to
AttributeError: 'NoneType' object has no attribute 'revision'
```
[requires]
openssl/1.1.1t
```

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,105 @@
...
+    path = conan_api.local.get_conanfile_path(args.path, cwd, py=None) if args.path else None
+    remotes = conan_api.remotes.list(args.remote) if not args.no_remote else []
+    overrides = eval(args.lockfile_overrides) if args.lockfile_overrides else None
+    lockfile = conan_api.lockfile.get_lockfile(lockfile=args.lockfile,
+                                               conanfile_path=path,
+                                               cwd=cwd,
+                                               partial=args.lockfile_partial,
+                                               overrides=overrides)
+    profile_host, profile_build = conan_api.profiles.get_profiles_from_args(args)
+    deps_graph = conan_api.graph.load_graph_consumer(path, args.name, args.version,
```
andreas-hilti, line 92: Maybe a bit more exotic (not sure whether we should support the following syntax):
conan recipe:create-sbom --requires openssl/1.1.1t
will lead to
```
  File "[...]\lib\site-packages\conan\api\subapi\graph.py", line 23, in _load_root_consumer_conanfile
    if path.endswith(".py"):
AttributeError: 'NoneType' object has no attribute 'endswith'
```
Note that the following command works fine:
conan graph info --requires openssl/1.1.1t

extensions/commands/recipe/cmd_create_sbom.py
andreas-hilti: This is fine for the (core) conan package. However, is this also as strict for conan-extensions?

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,16 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
...mplib.org"}], "licenses": [{"license": {"id": "GPL-2.0"}}, {"license": {"id": "LGPL-3.0"}}], "name": "gmp", "purl": "pkg:conan/gmp", "type": "library"}, "timestamp": "2023-08-08T12:55:48.275439+00:00", "tools": [{"externalReferences": [{"type": "build-system", "url": "https://github.com/CycloneDX/cyclonedx-python-lib/actions"}, {"type": "distribution", "url": "https://pypi.org/project/cyclonedx-python-lib/"}, {"type": "documentation", "url": "https://cyclonedx.github.io/cyclonedx-python-lib/"}, {"type": "issue-tracker", "url": "https://github.com/CycloneDX/cyclonedx-python-lib/issues"}, {"type": "license", "url": "https://github.com/CycloneDX/cyclonedx-python-lib/blob/main/LICENSE"}, {"type": "release-notes", "url": "https://github.com/CycloneDX/cyclonedx-python-lib/blob/main/CHANGELOG.md"}, {"type": "vcs", "url": "https://github.com/CycloneDX/cyclonedx-python-lib"}, {"type": "website", "url": "https://cyclonedx.org"}], "name": "cyclonedx-python-lib", "vendor": "CycloneDX", "version": "4.0.1"}]}, "serialNumber": "urn:uuid:b5b7a98b-e06a-4627-a520-2db2a4427daa", "version": 1, "$schema": "http://cyclonedx.org/schema/bom-1.4.schema.json", "bomFormat": "CycloneDX", "specVersion": "1.4"}
```
prince-chrismc, line 50: Is it possible to have a purl that's closers to conan's syntax?

instead of:
`"purl": "pkg:conan/m4@1.4.19?rref=c1c4b1ee919e34630bb9b50046253d3c",`

having something like:
`"purl": "pkg:conan~m4/1.4.19?#c1c4b1ee919e34630bb9b50046253d3c",`

I know of https://github.com/package-url/purl-spec but i've never used it so I apologize if it's a bad question :)

tests/test_create_sbom.py
```diff
@@ -0,0 +1,59 @@
...
+            def requirements(self):
+                self.requires("gmp/6.2.1")
+
+            def layout(self):
+                cmake_layout(self)
+
+            """)
+    save("conanfile.py", txt)
+
+    run("conan recipe:create-sbom . > sbom.json")
```
prince-chrismc, line 47: I like the general-ness of this command but I think having a much more specific `--format cyclonedx_json` would be more clear about the support 
https://docs.conan.io/2/reference/extensions/custom_commands.html?highlight=formatters#formatters

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: so if no 3rd party libs can be used, then the crafting of CycloneDX JSON, crafing of PURLs and other things need to be done by hand or re-implemented here, right? 

should he roll back, @prince-chrismc ?
well, if it would be not possible to use external libraries , then 

- crafting proper CycloneDX JSON is not that hard, as @hedtke did it before switching to the CycloneDX-lib via https://github.com/conan-io/conan-extensions/pull/66/commits/31b701ce9bc9f590a64a27830289deb006745bf6
- Crafting a valid CycloneDX license should be possible by checking whether a license is in the list of supported SPDX IDs (see <https://cyclonedx.org/schema/spdx.schema.json>). If the license is known, then [SPDX license ID](https://cyclonedx.org/docs/1.5/json/#tab-pane_components_items_licenses_oneOf_i0_items_license_oneOf_i0) can be used, else a [named license](https://cyclonedx.org/docs/1.5/json/#tab-pane_components_items_licenses_oneOf_i0_items_license_oneOf_i1) shall be used instead.
  Something like 
  ```py
  # a dict of all SPDX licenses supported by CycloneDX, keys are the values just lowercased - can also created dynamically like dict([k.lower(), k] for k in {... all values here ...})
  _cdx_licenses_lowercased = {'0bsd':'0BSD', 'aal':'AAL', '...': '... and all others ...'}  # all the values from https://cyclonedx.org/schema/spdx.schema.json 

  make_cdx_license(license: str) -> dict: 
    licenseId = _cdx_licenses_lowercased.get(license.lower(), None)
    return {'id':licenseId} if licenseId else {'name':licenseId}
  ```
- Crafting a PURL is not that hard either, it just needs carefully url-encoding several parts of the string, as described here https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component and here https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#character-encoding

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,105 @@
...
+            type=ExternalReferenceType.WEBSITE,
+            url=XsUri(n.conanfile.homepage),
+        ))
+    return result
+
+
+@conan_command(group="Recipe")
+def create_sbom(conan_api: ConanAPI, parser, *args):
+    """
+    creates an SBOM in CycloneDX 1.4 JSON format from a Conan graph JSON
```
andreas-hilti, line 75: The "graph JSON" part is outdated by now.

extensions/commands/recipe/cmd_create_sbom.py
hedtke: @prince-chrismc really no external libs in extensions, too?

extensions/commands/recipe/README.md
hedtke: I don't think so, the definition is very strict about how to add qualifiers (rref) and about the delimiter between attributes, like `pkg:type/name@version...`

extensions/commands/recipe/cmd_create_sbom.py
hedtke: None value checked and test added

extensions/commands/recipe/cmd_create_sbom.py
hedtke: None value checked and test added

extensions/commands/recipe/cmd_create_sbom.py
hedtke: dropped

extensions/commands/recipe/README.md
hedtke: I now enforce a format option.

If we allow the usage of external libraries, this would enable us to also output the SBOM in CycloneDX JSON1.2, JSON1.3, JSON1.5, XML1.0, XML1.1, XML1.2, XML1.3, XML1.4, and XML1.5

If I have to drop all external libs and create the output by hand, I will stick to one of them only, probably JSON1.4

tests/test_create_sbom.py
hedtke: Added

hedtke:
Ready for next review round. I think I resolved all issues. The only remaining point I see at the moment is whether external libs are forbidden. If so, I have to rewrite lots of things by hand. Thank you @jkowalleck for the detailed points. If we can allow libs, it would support more output formats and we should be almost done.

extensions/commands/recipe/cmd_create_sbom.py
memsharded: At the moment, the ``conan-extensions`` doesn't have a mechanism to install ``pip`` dependencies in the current Python environment automatically, so at the moment this should be something done when installing ``pip install conan package-url ...`` as necessary. 

Note that adding them to a ``requirements.txt`` in this current repo is pointless, as things here are intended to be installed with ``conan config install``.

We can think in the future of possible ways to automate this, at the moment I'd recommend to ``try-except`` the import with a clear message ``The sbom extension needs the 'package-url' package, please 'pip install ...'``

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,134 @@
...
+from cyclonedx.model import ExternalReference, ExternalReferenceType, LicenseChoice, XsUri
+from cyclonedx.model.bom import Bom
+from cyclonedx.model.component import Component, ComponentType
+from cyclonedx.output.json import JsonV1Dot4
+
+from conan.api.output import cli_out_write, ConanOutput
+from conan.api.conan_api import ConanAPI
+from conan.cli.args import common_graph_args, validate_common_graph_args
+from conan.cli.command import conan_command
+from conans.client.graph.graph import Node
```
memsharded, line 14: This ``from conans`` import is risky, will break in future releases, I'd avoid it, specially if it is only for types

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,134 @@
...
+        deps_graph = conan_api.graph.load_graph_consumer(path, args.name, args.version,
+                                                     args.user, args.channel,
+                                                     profile_host, profile_build, lockfile,
+                                                     remotes, args.update)
+    else:
+        deps_graph = conan_api.graph.load_graph_requires(args.requires, args.tool_requires,
+                                                         profile_host, profile_build, lockfile,
+                                                         remotes, args.update)
+    # END COPY
+    globals()["unknown_name_created"] = False
```
memsharded, line 126: Is this ``globals`` being used elswhere, beyond the above usage in L42? If only in L42, it would be better to pass it by parameter. Formatters can take any number of arguments, just pass a ``dict`` as result of the command, and let the formatters take things from the dic.

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,134 @@
...
+    serialized_json = JsonV1Dot4(result).output_as_string()
+    cli_out_write(serialized_json)
+
+
+def format_text(result):
+    out = ConanOutput()
+    out.error("Format 'text' not supported")
+
+
+@conan_command(group="Recipe", formatters={
```
memsharded, line 94: The ``conan recipe:create_sbom`` reads a bit generic. (the ``recipe`` layer reads generic). I think it could make sense to have something like "sbom", "cyclonedx" as the layer name.

Not a big deal, we can move forward, keep the discussion and iterate these things, doesn't need to be agreed or changed right now.

memsharded (commented):
Thanks very much for all the effort. This has gone a long way since the initial PR and also thanks to the reviewers and the feedback, I really think that this is now in a much better place than the original effort. I think we might need some iteration for naming and align some Conan general practices, but overall looks good.

memsharded:
I have launched the tests, seems CI is not passing: https://github.com/conan-io/conan-extensions/actions/runs/5806921433/job/15741930416?pr=66

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: > I'd avoid it, specially if it is only for types

if it was for typing only, then you could follow [pep848 section "Runtime or type checking?"](https://peps.python.org/pep-0484/#runtime-or-type-checking)

hedtke: I am using the import now only during type checking

extensions/commands/recipe/cmd_create_sbom.py
hedtke: Global is replaced by a dict now

hedtke:
Ready for the next round. For now, we stick to using the external libs but import them with try/except and notify the user when something is missing and how to install the required libs.

Tests run on my local machine.

I am fine with changing the name/layer. If we keep the external libs, maybe something like
```
conan sbom:cyclonedx --version 1.4 --format xml ...
```
This would allow later for alternative SBOM standards, and for CycloneDX we could offer all versions in JSON and XML format. What do you think?

extensions/commands/recipe/cmd_create_sbom.py
hedtke: I found a problem with that approach: In installed the extensions in an environment where I do not have the external libs installed. Now even running a non-extension command triggers an Exception:

```
$ conan install .
Migration: Successfully updated settings.yml
The sbom extension needs the 'packageurl' package, please run 'pip install packageurl-python'
ERROR: Exiting with code: 1
```

hedtke: I changed the approach: Required libs are checked only then the new command is called

.github/workflows/test_conan_extensions.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install -U pip
-        pip install pytest
+        pip install pytest packageurl-python cyclonedx-python-lib
```
jkowalleck, line 21: please add version constraints, like 
`packageurl-python>=0.10,<0.11`

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,32 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+#### [Create SBOM](cmd_create_sbom.py)
+
+Creates an SBOM in CycloneDX 1.4 JSON format.
+
+**Parameters**
+* **format** _Required_: `cyclonedx_1.4_json`
+* supports all other arguments used by `conan graph`, see `conan recipe:create-sbom --help`
+
+**Dependencies**
+The command requires the packages `packageurl` and `cyclonedx`.
```
jkowalleck, line 49: do not depend on `cyclonedx`, this is not a library but a set of executables.
depend on `cyclonedx-python-lib` instead.

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,32 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+You can install them via
+
+```bash
+pip install packageurl-python cyclonedx-python-lib
```
jkowalleck, line 53: please add version constraints, like 
`packageurl-python>=0.10,<0.11`

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,32 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+```
+
+Usage:
+
+```
```
jkowalleck, line 58: please add code fence language `shellSession`
like so
```text
```shellSession
$ cd conan-center-index/recipes/gmp/all
$ conan recipe:create-sbom -f cyclonedx_1.4_json .
{"components": [{"bom-ref": 
...
````

File: extensions/commands/recipe/README.md [hunk 21]
```diff
@@ -36,3 +36,32 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+
+```
```
jkowalleck, line 64: please add code fence language `shellSession`
like so
```text
```shellSession
$ conan recipe:create-sbom -f cyclonedx_1.4_json --requires gmp/6.2.1
{"components": [{"bom-ref": 
...
````

File: extensions/commands/recipe/cmd_create_sbom.py [hunk 22]
```diff
@@ -0,0 +1,149 @@
...
+    "cyclonedx_1.4_json": format_cyclonedx_14_json})
+def create_sbom(conan_api: ConanAPI, parser, *args):
+    """
+    creates an SBOM in CycloneDX 1.4 JSON format
+    """
+    for package_name, found in found_libs.items():
+        if not found:
+            sys.stderr.write(
+                f"The sbom extension needs an additional package, please run 'pip install {package_name}'\n")
+            sys.exit(1)
```
memsharded, line 116: It is a simpler to make the imports with try-except local. It is not compulsory that imports are in the global scope, they can be local to functions.

memsharded (approved):
Almost there!
I think incorporating the reviews from @jkowalleck we would be ready for initial merge (important to consider it experimental at this stage, to allow breaking changes in future iterations) to start gathering feedback.

extensions/commands/recipe/cmd_create_sbom.py
hedtke: imports are local now

extensions/commands/recipe/README.md
hedtke: Language is now specified

extensions/commands/recipe/README.md
hedtke: Language is now specified

extensions/commands/recipe/README.md
hedtke: version constraint added

.github/workflows/test_conan_extensions.yml
hedtke: added

extensions/commands/recipe/cmd_create_sbom.py
hedtke: dict is replaced by a static variable

hedtke:
Ready for next review round. Thank you all for the feedback!

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,33 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+The command requires the package `cyclonedx-python-lib`.
+You can install it via
+
+```shell
+$ pip install "cyclonedx-python-lib>=4.0.1,<5.0.0"
+```
+
+Usage:
+
+```shell
```
jkowalleck, line 59: change:
````patch
- ```shell
+ ```shellSession
````

File: extensions/commands/recipe/README.md [hunk 24]
```diff
@@ -36,3 +36,33 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+```
+
+```shell
```
jkowalleck, line 65: change:
````patch
- ```shell
+ ```shellSession
````

File: extensions/commands/recipe/cmd_create_sbom.py [hunk 25]
```diff
@@ -0,0 +1,141 @@
...
+    """
+    see https://cyclonedx.org/docs/1.4/json/#components_items_licenses
+    """
+    from cyclonedx.factory.license import LicenseFactory
+    from cyclonedx.model import LicenseChoice
+    if ids is None:
+        return None
+    if not isinstance(ids, tuple):
+        ids = [ids]
+    return [LicenseChoice(license=LicenseFactory().make_from_string(i)) for i in ids]
```
jkowalleck, line 32: I know of reports telling that multiple licenses added in the CycloneDX library caused invalid results.
You should use only one license, if you can.

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,141 @@
...
+                                                         remotes, args.update)
+    else:
+        deps_graph = conan_api.graph.load_graph_requires(args.requires, args.tool_requires,
+                                                         profile_host, profile_build, lockfile,
+                                                         remotes, args.update)
+    # END COPY
+    name.unknown_name_created = False
+    components = {n: create_component(n) for n in deps_graph.nodes}
+    bom = Bom()
+    bom.metadata.component = components[deps_graph.root]
```
jkowalleck, line 136: please add your own to `bom.metadata.tools`, so people know how the SBOM was built.
remmeber: `bom.metadata.tools` is prepoppulated, so just use `bom.metadata.tools.add(Tool(...))`

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,141 @@
...
+                                                         args.user, args.channel,
+                                                         profile_host, profile_build, lockfile,
+                                                         remotes, args.update)
+    else:
+        deps_graph = conan_api.graph.load_graph_requires(args.requires, args.tool_requires,
+                                                         profile_host, profile_build, lockfile,
+                                                         remotes, args.update)
+    # END COPY
+    name.unknown_name_created = False
+    components = {n: create_component(n) for n in deps_graph.nodes}
```
jkowalleck, line 134: maybe use another way to define the dict's keys. an object for a key might lead to runtime-dependent issues.

extensions/commands/recipe/README.md
hedtke: Changed

extensions/commands/recipe/README.md
hedtke: Changed

extensions/commands/recipe/cmd_create_sbom.py
hedtke: What do you have in mind? The only thing I can think of is `id(n)`, which is a bit ugly

extensions/commands/recipe/cmd_create_sbom.py
hedtke: Added

extensions/commands/recipe/cmd_create_sbom.py
hedtke: I am not sure how to tackle that. The CycloneDX standard allows multiple licenses. It is supported by the library. The generated JSON is correct and dependencytrack for example has no problem parsing it (according to the log). What should I do? If I pick only one, which one?

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: > A dictionary’s keys are almost arbitrary values. Values that are not hashable, that is, values containing lists, dictionaries or other mutable types (that are compared by value rather than by object identity) may not be used as keys.

So, as long as the has does not change, you are fine, i guess.
I am no expert on this specific subject here.

did some tests, objects for keys seams fine in py3.7
test script
```py
o1 = object()
o2 = object()
d = {o1:1, o2:2}
for e in d.keys():
  print(type(e), repr(e))
```

extensions/commands/recipe/cmd_create_sbom.py
jkowalleck: do you have an example of a Conan package with multiple licenses?

hedtke: bliss has two: https://conan.io/center/recipes/bliss?version=0.77
https://github.com/conan-io/conan-center-index/blob/master/recipes/bliss/all/conanfile.py#L17

jkowalleck: could you try a project with this `bliss` dependency and upload the CycloneDX result as a gist? 
https://gist.github.com/

i really love to see whether everything results in a valid result for this case.
as i mentioned, there were reports that multi-licenses were broken in the library, but i do not remember seeing any examples of an issue there. (despite the code shows evidence that something might be off)

hedtke: https://gist.github.com/hedtke/f7f2866fbd80504a6d3bae97fc61ad92

jkowalleck: > https://gist.github.com/hedtke/f7f2866fbd80504a6d3bae97fc61ad92

fantastic. than you so much. relevant part:
https://gist.github.com/hedtke/f7f2866fbd80504a6d3bae97fc61ad92#file-sbom-json-L33-L44

This s just as expected:
- multiple licenses in the document :+1: - so the library has none of the reported issues
- all licenses are names, as they do not match any of the well-known SPCDX licenses,
  according to https://github.com/CycloneDX/specification/blob/master/schema/spdx.schema.json
  compared to https://spdx.org/licenses/

----

**conclusion:** you are golden.

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,141 @@
...
+
+def package_type_to_component_type(pt: str):
+    from cyclonedx.model.component import ComponentType
+    if pt is "application":
+        return ComponentType.APPLICATION
+    else:
+        return ComponentType.LIBRARY
+
+
+def licenses(ids):
```
jkowalleck, line 22: maybe add typing, as later it is checked for `isinstance(ids, tuple)`
```py
from typing import Iterable, Optional, Tuple, Union
def licenses(ids: Union[Tuple[str], str]) -> Optional[Iterable['LicenseChoice']]:
  ...
```

hedtke: Type annotation added, but I cannot use `'LicenseChoice'` as this is not imported at that point. The import happens inside the body of the method.

jkowalleck: i will pullrequested some restructuring, that you can simply merge into your PR:
then, all should work fine and allow a much easier maintenance.

PR coming soon.

jkowalleck:
@hedtke I'd propose some changes via https://github.com/hedtke/conan-extensions/pull/1
If my PR was merged into yours, then my changes would apply in this very PR here.

hedtke:
Thanks @jkowalleck for the restructuring.
Ready for next review round

.gitignore
```diff
@@ -1,3 +1,5 @@
+/tests/tmp*
```
jkowalleck, line 1: locally running pytest might create these

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -1,122 +1,110 @@
...
+        from cyclonedx.model.bom import Bom
+        from cyclonedx.model.component import Component, ComponentType
+        from packageurl import PackageURL
     except ModuleNotFoundError:
-        sys.stderr.write(
-            "The sbom extension needs an additional package, please run 'pip install cyclonedx-python-lib'\n")
+        # Assert on RUNTIME of the actual conan-command, that all requirements exist.
+        # Since conan loads all extensions when started, this check could prevent conan from running,
+        # if loading dependencies is performed outside the actual conan-command in global/module scope.
+        print('The sbom extension needs an additional package, please run:'
```
jkowalleck, line 43: @hedtke I forgot a comma(`,`) at the end of this line, right after `please run:'`
Could you add it for me? 

my idea was, that every paras of `print()` would be a new line, ths is why the `, sep='\n'` is at the end

jkowalleck: THanks.

this was done via ee19212cfecf295888c83eea8c6a36cdda81a7d8

jkowalleck (approved):

jkowalleck:
> Ready for the next round. For now, we stick to using the external libs but import them with try/except and notify the user when something is missing and how to install the required libs.
> 
> Tests run on my local machine.
> 
> I am fine with changing the name/layer. If we keep the external libs, maybe something like
> 
> ```
> conan sbom:cyclonedx --version 1.4 --format xml ...
> ```
> 
> This would allow later for alternative SBOM standards, and for CycloneDX we could offer all versions in JSON and XML format. What do you think?

From a programming perspective, the following should be a low hanging fruit, and it would be future prove for future/other sbom standards

```sh
conan sbom:cyclonedx --format xml1.4 ...
```

hedtke:
I added a commit (that we can revert if we want to) that supports all output formats. From the docu:

**Parameters**
* `sbom_format` _Required_:  `1.4_json`, `1.3_json`, `1.2_json`, `1.4_xml`, `1.3_xml`, `1.2_xml`, `1.1_xml`, or `1.0_xml`
* supports all other arguments used by `conan graph`, see `conan recipe:create-sbom --help`

extensions/commands/recipe/cmd_create_sbom.py
```diff
@@ -0,0 +1,145 @@
...
+
+    components = {node: create_component(node) for node in deps_graph.nodes}
+    bom = Bom()
+    bom.metadata.component = components[deps_graph.root]
+    bom.metadata.tools.add(me_as_tool())
+    for node in deps_graph.nodes[1:]:  # node 0 is the root
+        bom.components.add(components[node])
+    for dep in deps_graph.nodes:
+        bom.register_dependency(components[dep], [components[dep_dep.dst] for dep_dep in dep.dependencies])
+    return {'bom': bom, 'sbom_format': args.sbom_format, 'formatters': formatters}
```
andreas-hilti, line 145: Are the different formats working? I might be doing something wrong, but
```
conan recipe:create-sbom --sbom_format 1.3_json .
```
results for me in
```
[...] "specVersion": "1.4"}
```

tests/test_create_sbom.py
```diff
@@ -0,0 +1,83 @@
...
+    (str(), "doesnotmatter.txt", f"--requires {REQ_LIB}/{REQ_VER}", False)
+])
+def test_create_sbom_cdx14json(conanfile_content, conanfile_name, sbom_command, test_metadata_name):
+    repo = os.path.join(os.path.dirname(__file__), "..")
+    run(f"conan config install {repo}")
+    run("conan profile detect")
+    save(conanfile_name, conanfile_content)
+
+    # discard all STD_ERR(2)
+    out = run(f"conan recipe:create-sbom --sbom_format 1.4_json {sbom_command} 2>/dev/null")
```
andreas-hilti, line 81: I'm not familiar with tools.run, but is this portable?
- redirect: >2
- /dev/null

extensions/commands/recipe/cmd_create_sbom.py
andreas-hilti: I noticed that the xml versions are working; is this an issue in cyclonedx-python-lib?

hedtke: Copy and paste error. I will fix it

tests/test_create_sbom.py
andreas-hilti: At least for me (on Windows), the tests pass only if I change it to "2>output.err".

extensions/commands/recipe/cmd_create_sbom.py
andreas-hilti: Should have seen that myself as well...

tests/test_create_sbom.py
hedtke: I extended the run method to deal with that

extensions/commands/recipe/cmd_create_sbom.py
hedtke: Fixed

tests/tools.py
```diff
@@ -1,15 +1,15 @@
 import subprocess
 
 
-def run(cmd, error=False):
+def run(cmd, error=False, stderr=subprocess.PIPE):
```
jkowalleck, line 4: all these changes, just because i wrote a test that was intended to drop  out/errormessages?

then please have if changed like this: 

```patch
- def run(cmd, error=False, stderr=subprocess.PIPE):
8 def run(cmd, error=False, *, stdout=subprocess.PIPE, stderr=subprocess.PIPE):
```

this way each stream can be dopped/piped individually,
and the `*` causes the downstream users to be forced to name the param, like 
```
run('echo foo', out=subprocess.PIPE)



-----



but all of this was not needed, if this function could simply return stdout and stderr separately, as a tuple.

```py
def run(...)>
  return out, err

_, err = run('echo 1 && foobar')
out, _ = run('echo 1') 
out, err = run('echo 1 && foobar')
```

hedtke: stdout is now also a parameter

tests/test_create_sbom.py
```diff
@@ -0,0 +1,82 @@
...
+
+            """)
+
+
+@pytest.mark.parametrize("conanfile_content,conanfile_name,sbom_command,test_metadata_name", [
+    (create_conanfile_py(), "conanfile.py", ".", True),
+    (create_conanfile_txt(), "conanfile.txt", ".", False),
+    (str(), "doesnotmatter.txt", f"--requires {REQ_LIB}/{REQ_VER}", False)
+])
+def test_create_sbom_cdx14json(conanfile_content, conanfile_name, sbom_command, test_metadata_name):
```
jkowalleck, line 74: could you add test for all supported formats?

hedtke: Added

prince-chrismc:
I ran the tests and they are failing to find the version of cycloneDX

hedtke:
I didn't realize Python 3.6 and 3.7 are still supported although they do not even get security fixes anymore. I will give and older cyclone version a try and check whether that works

hedtke:
I added support for cyclonedx-python-lib 3 which is the latest major version supported for python 3.6

hedtke:
Not easy to get everything right. I found a way to differentiate between cyclonedx v3 (for Python 3.6) and v4 (for later) without using `importlib.metadata` (since Python 3.8) and without `cyclonedx.__version__`, see https://github.com/CycloneDX/cyclonedx-python-lib/issues/416

prince-chrismc:
Unfortunately I do not have the knowledge to review this at a technical level. At a high level this looks really good with my background using them in the past. I like the CLI for this is the most intuitive so far, I think it's a great start 👏  I also like the pattern for using extern deps + the error message - very well done 👍  

As for the code, and getting this merged we'll need to wait for someone of the team to make that decision :)

jkowalleck:
> I also like the pattern for using extern deps + the error message - very well done

Thanks. I think this is a pattern that extensions could use.
I encapsulated almost **all** the code into the extension's runner function, including the imports. This way nothing is imported or created when the extension is loaded. 
Only when the extension is executed, then the needed externals are probed.

extensions/commands/recipe/README.md
```diff
@@ -36,3 +36,33 @@ updating moltenvk/1.2.0 to moltenvk/1.2.2 in conan-center-index/recipes/qt/6.x.x
...
+<!-- keep in sync with the error message that asks for requirements/dependencies in `cmd_create_sbom.py` -->
+```shellSession
+$ pip install 'cyclonedx-python-lib>=3.1.5,<5.0.0'
+```
+
+Usage:
+
+```shellSession
+$ cd conan-center-index/recipes/gmp/all
+$ conan recipe:create-sbom --sbom_format 1.4_json .
```
memsharded, line 61: Why ``--sbom_format`` and not the generic ``--format`` common to all Conan commands?

Maybe to be able to embed the ``imports`` in the command?

memsharded, line 61: I think we still need to look for a better "layer" name than "recipe"

memsharded (approved):
I think this is ready to be merged from your side, it just needs a first quick round of internal testing and feedback from our side.

We will play a little bit with it, probably try to refactor a bit the code to be more aligned with some of our style, and consider and think about some changes like the "layer" ("recipe") name. Sorry we haven't managed to do this yet, we have been swamped, but we haven't forgot about this PR. Thanks!

extensions/commands/recipe/README.md
jkowalleck: actually there was a pythonic and pretty implementation for this,
at state 4ce29c024ebd587ce8a40f0c970b298f20ac6560

see https://github.com/conan-io/conan-extensions/blob/4ce29c024ebd587ce8a40f0c970b298f20ac6560/extensions/commands/recipe/cmd_create_sbom.py#L16-L29


but it got changed to what it is now ... and not for the better, i would say.

jkowalleck: actually there was a pythonic and pretty implementation for this,
at state 4ce29c024ebd587ce8a40f0c970b298f20ac6560

see https://github.com/conan-io/conan-extensions/blob/4ce29c024ebd587ce8a40f0c970b298f20ac6560/extensions/commands/recipe/cmd_create_sbom.py#L16-L29


but it got changed to what it is now ... and not for the better, i would say.

jkowalleck: i'd suggest `conan sbom:cyclonedx --format 1.4_xml` as stated here: https://github.com/conan-io/conan-extensions/pull/66#issuecomment-1678740469

hedtke: OK, I changed the name of the command to `conan sbom:cyclonedx --format 1.4_xml`

hedtke:
Ready for the next review round

extensions/commands/sbom/cmd_cyclonedx.py
```diff
@@ -0,0 +1,183 @@
...
+
+def format_text(_: 'Bom') -> None:
+    supported = ', '.join([v for v in formatter.keys() if v is not 'text'])
+    raise ConanException(f"Format 'text' not supported. Supported values are: {supported}")
+
+
+formatter["text"] = format_text
+
+
+@conan_command(group="Recipe", formatters=formatter)
```
jkowalleck, line 44: the `group` does not match the path.
shold it not be `SBOM` ?

hedtke: Fixed

extensions/commands/sbom/cmd_cyclonedx.py
```diff
@@ -0,0 +1,183 @@
...
+                url=XsUri("https://github.com/conan-io/conan-extensions")))  # noqa
+        else:
+            tool.external_references.add(ExternalReference(
+                reference_type=ExternalReferenceType.WEBSITE,
+                url=XsUri("https://github.com/conan-io/conan-extensions")))
+        return tool
+
+    # region COPY FROM conan: cli/commands/graph.py
+    common_graph_args(parser)
+    args = parser.parse_args(*args)
```
memsharded, line 151:
```suggestion
    # FIXME: Process the ``--build-require`` argument
    parser.add_argument("--build-require", action='store_true', default=False,
                        help='Whether the provided path is a build-require')
    args = parser.parse_args(*args)
```

This was added to the ``validate_common_graph_args()`` checks recently, but the ``common_graph_args()`` doesn't add it yet. It will probably do it, but this is part of the stabilization of the Python API for custom commands. It is expected that this will still break.

memsharded:
Ok, so this PR is merged, thanks very much all @hedtke for the contributions and all the reviewers!

This is a great first step. We still have some work ahead, while the APIs stabilize, there will probably be breakages that we will need to fix. We would like also take the chance of future fixes and other tasks to probably go over the code and propose some refactors or style changes to align a bit better with the rest of the Conan codebases, but that shouldn't be very relevant.

memsharded:
For new features, command args changes, etc (anything beyond bugfixes), I'd recommend submitting issues, discussing there the design, etc, and then move to PRs after alignement.

PR #80, opened 2023-10-16
Add author to SBOM components

erik-moqvist:


danimtb:
Hi @erik-moqvist, thanks for the contribution.
Could you elaborate a little bit about this fix? What is the issue or motivation to add the author?

erik-moqvist:
The minimum elements document mentions author in a few places, so I figured it's probably a good idea to include them. It's not 100% clear to me which author the document wants, but I have seen other tools add authors of dependencies.

https://www.ntia.doc.gov/files/ntia/publications/sbom_minimum_elements_report.pdf

danimtb:
Hi @erik-moqvist,

Thanks a lot for the context and the linked document.
I have checked it and it seems that the author field in the document references the author of the SBOM document, not the author of the software. However, I am not sure about this in the case of the author of each component, which indeed makes sense for it to be the "author" of the recipe or even better, the authot of the library.

I will be taking a deeper look into it. Thanks a lot!

danimtb:
Indeed:

> Author of SBOM Data The name of the entity that creates the SBOM data for this
> component.

The author data field for each component should be the entity that is creating the SBOM, so it seems that using the conanfile author is not the desired approach according to the document. Maybe the way to go is to add a flag to the custom command to allow adding a custom author to the SBOM. WDYT?

erik-moqvist:
Maybe the cyclonedx JSON specification is a better reference: https://cyclonedx.org/docs/1.4/json/#components_items_author

It says, author for a component is: "The person(s) or organization(s) that authored the component"

danimtb:
ok, that makes more sense. Do you think this should be the author of the recipe or the author of the software? Probably getting the info in the last case is not so easy... maybe it is just goog as it is

danimtb (approved):

danimtb:
@erik-moqvist thanks a lot for the contribution! I'll open a PR with a default "conan" author in case the recipe does not have the author attribute (assuming the recipe is coming from conan-center, as recipes there do not have this attribute set)

PR #84, opened 2023-10-31
Better handling of errors when creating Build Info

czoido:


extensions/commands/art/utils.py
```diff
@@ -28,6 +40,26 @@ def response_to_str(response):
     except Exception:
         return response.content
 
+class UnauthorizedException(ConanException):
+    """Exception raised for unauthorized requests (HTTP 401)."""
+    pass
+
+class ForbiddenException(ConanException):
+    """Exception raised for forbidden requests (HTTP 403)."""
+    pass
```
memsharded, line 49: Some of these exceptions are duplicated from Conan client. Why ``ConanException`` is reused from the client and these are duplicated?

czoido: I did not want to import things from Conan that are not public, the only public ones are `ConanException`, `ConanInvalidConfiguration` and `ConanMigrationError`

memsharded: Ah, ok, makes sense! 👍

PR #90, opened 2023-11-14
Fix for when args.path is not used

michxymi:
With the current implementation of the CCI `export_all_versions` command you can either pass a config file, a path or a name.

There's currently a bug when not using a path argument. When either the config file or a name is parsed `args.path` is still used to locate the recipes directory. However because no path has been parsed as an arg it is undefined (i.e None).

This PR adds a check condition to check if `args.path` is undefined and if it is, it uses the `cwd/recipes` path, to comply with the expectation that this script should run from the root of a CCI file tree. When a path is indeed requested and `args.path` is not undefined, its value is used.

### DIFF
--- a/extensions/commands/cci/cmd_export_all_versions.py
+++ b/extensions/commands/cci/cmd_export_all_versions.py
@@ -101,7 +101,7 @@ def export_all_versions(conan_api, parser, *args):
         folders_to_export = item[recipe_name][0]['folders'] if isinstance(item, dict) else None
         out.verbose(f"Processing recipe '{recipe_name}'")
 
-        recipe_folder = os.path.join(args.path, recipe_name)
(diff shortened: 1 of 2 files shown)


michxymi:
@uilianries Decided to make my first contribution :)

uilianries (approved):
@michxymi Thank you for your first contribution! 🎉

RubenRBS (approved):
Thanks a lot! :)

We're delaying the merge until we do some behind the scenes work on the repo itself, will ping you when ready

michxymi:
Thanks both. I should have started from this but I also added some unit tests for all the possible args.

uilianries (approved):
New tests!! ❤️

RubenRBS (approved):
This got lost in the sea of other work - thanks to @danimtb for the ping that this was still open :)

Issue #60, opened 2023-06-19
[question] How to package an application with its dependencies libraries?


Todiq:
### What is your question?

Hello,

This is to get more clarification regarding conan-io/conan#14067

I ran your code @RubenRBS:
```python
from conan.tools.files import copy

def deploy(graph, output_folder, **kwargs):
    for node in graph.nodes:
        libdirs = node.conanfile.cpp_info.libdirs
        if libdirs:
            for libdir in libdirs: 
                copy(node.conanfile, "*.so*", src=libdir, dst="your/final/path")
```

Two things were to be fixed. The first one was the import (```files``` instead of ```file```). The second one was to change the copy from ```*.so``` to ```*.so*``` to prevent some links to be broken, like for ```boost``` libraries.

I have a few questions:
- Is it possible to run a ```cmake install``` at the beginning of the ```deploy()``` method? I already do so inside my ```package()```method:
```python
def package(self):
	cmake = CMake(self)
	cmake.install()
```
but I do not know if I can call that one directly.

- Is it possible to remove some of the unrequired libs? In fact, taking the example of boost once again, one does not necessarily needs all the libs, since it only makes the deployment folder bigger.

- Is it possible to add system libs? I need to add python to my folder, but since the recipe of cpython is not Conan 2.X ready just yet on conancenter, I am forced to link with the system Python in my CMakeLists.txt.

All of these questions serve one purpose : I want to run a single conan command to package (build done beforehand) and create a zip file of my project and all its dependencies. I then want to distribute it to clients (they will have the proper environment to run it, but won't have conan installed). Do you think you could give me a hint?

Thanks in advance.

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

Todiq:
> https://github.com/conan-io/conan/pull/15382 merged, this will be in next Conan 2.5

Thanks!

PR #94, opened 2023-12-08
Add dll_deploy.py

fdgStilla:
Since conan import was removed from conan 2, this deployer try to replace this feature by copying the dll from the dependencies to the output folder.

Related to https://github.com/conan-io/conan-extensions/issues/60

Some notes:
- tested only locally with conan  2.0.14
- shall we skip the dll if `require.build` is True?
- for now it is copied to the output/bin directory, in most cases it will allow to copy the dll next to the binary without using the `--deployer-folder` argument

### DIFF
--- a/extensions/deployers/dll_deploy.py
+++ b/extensions/deployers/dll_deploy.py
@@ -1,0 +1,28 @@
+import os, shutil
+
+def deploy(graph, output_folder, **kwargs):
+    conanfile = graph.root.conanfile
+    conanfile.output.info(f"Deploying the dll...")
+    for req, dep in conanfile.dependencies.items():
+        conanfile.output.verbose(f"Searching for dll in {dep.ref}...")
+        if dep.package_folder is None:
+            conanfile.output.verbose(f"{dep.ref} does not have any package folder")
+            continue
+        if not dep.cpp_info.bindirs:
+            conanfile.output.verbose(f"{dep.ref} does not have any bin directory")
+            continue
+        for bindir_name in dep.cpp_info.bindirs:
+            bindir_path = os.path.join(dep.package_folder, bindir_name)
+            if not os.path.isdir(bindir_path):
+                conanfile.output.verbose(f"{bindir_path} does not exist")
+                continue
+            file_count = 0
+            for file_name in os.listdir(bindir_path):
+                if file_name.endswith(".dll"):
+                    file_count += 1
+                    file_path = os.path.join(bindir_path, file_name)
+                    dst = os.path.join(output_folder, "bin", file_name)
+                    conanfile.output.verbose(f"Copy {file_path} in {dst}")
+                    shutil.copy2(file_path, dst)
+            conanfile.output.info(f"Copied {file_count} dll from {dep.ref}")
+    conanfile.output.info(f"Deployed all dll!")


memsharded:
Hi @fdgStilla 

Thanks very much for your contribution.
I think, and I have been wanting to have a built-in deployer in conan for runtime (dll, exes). So I think it would be good to move this to the conan repo, and try to contribute this as a built-in:

- Lets call it ``runtime_deployer``
- It should target all platform, collecting dlls, dylibs, so and executables (basically the ``bin`` folder plus some ``lib`` artifacts)
- It should collect them "flat", all in the same folder
- I think it should also collect licenses, in different folders so they don't overwrite

What do you think? Does this align with your goal?

fdgStilla:
My high level purposes for this deployer are:
- be able to launch the executable on windows, without using the `conanrun.sh` script (so that you can use the Visual Studio IDE with cmake support more easily)
- be able to create an installer, by copying all the .dll into a single folder, then we can use any installer tool very easily (qt installer, innosetup etc)

> It should target all platform, collecting dlls, dylibs, so and executables (basically the bin folder plus some lib artifacts)
- we don't have any apple equipment I would not be able to test this part
- on windows .dll in the same directory as the binary are automatically loaded, but .so files are not (you have to play with LD_LIBRARY_PATH, rpath or other) so it is not really useful to have them next to the binary, but why not for creating an installer
- copying executables would not fit with my purpose of creating an installer

> I think it should also collect licenses, in different folders so they don't overwrite
I understand the need for this deployment but I think it shall be a different deployer.

memsharded:
> be able to launch the executable on windows, without using the conanrun.sh script (so that you can use the Visual Studio IDE with cmake support more easily)

For the first goal, you should track: https://github.com/conan-io/conan/pull/15192. This might be way more convenient than deploying dlls

> be able to create an installer, by copying all the .dll into a single folder, then we can use any installer tool very easily (qt installer, innosetup etc)

Yes, that is one of the main use cases

> we don't have any apple equipment I would not be able to test this part

It is fine, that can be contributed by others

> on windows .dll in the same directory as the binary are automatically loaded, but .so files are not (you have to play with LD_LIBRARY_PATH, rpath or other) so it is not really useful to have them next to the binary, but why not for creating an installer

This is still useful, as can be solved with a simple, single LD_LIBRARY_PATH="." definition. And also useful for installers or re-packaging in other formats (deb, rpm)

> copying executables would not fit with my purpose of creating an installer

Why not? Do they really affect you? Maybe you can just remove them as the first step of your installer creation?


I'd say that we would still want to move this forward as built-in, generic ``runtime_deployer``, and not having a very specific just for windows, just dlls, but not exes deployers. Maintaining the built-in one will serve for the majority of Conan users, and for the cases that someone wants to not deploy the exes, it would be a custom deployer just commenting out the line copying the executables.

fdgStilla:
> For the first goal, you should track: https://github.com/conan-io/conan/pull/15192. This might be way more convenient than deploying dlls

Good to know! Unfortunatly we are stuck with CMakeSettings.json for some projects and it would be too complicated to explain our internal technical challenges.

> Why not? Do they really affect you? Maybe you can just remove them as the first step of your installer creation?

Then we would have to identify the exe produced by the project and the exe produced by the conan install when invoking the installer. There are of course so many ways to solve this but I am just highlighting the fact it is more complicated than just copy the dll like we used to do with conan import.

But I understand you point of having a generic runtime installer, how do you want me to contribute?

memsharded:
> But I understand you point of having a generic runtime installer, how do you want me to contribute?

I think we can close this ticket. Then the process could be:

- Submit the PR with this generator, as-is to the Conan project, lets call it ``runtime_deploy``
- Need a small test. I think it can be one, with a fake ".dll", no need to be a compiled one. You can start by copying the ``test_builtin_full_deploy`` test in "test_install_deploy.py", or if you need any help here, just let us know and we can contribute the test.

From there we will take it, probably first integrating that without documenting or making it public, then adding more support for Linux and Mac, and then after some stabilization, making it public.

PR #96, opened 2023-12-19
[command] Update Status Page via Conan extensions

uilianries:
This new Conan extension command will add the action of updating Atlassian Status page via Conan extensions. 

So far, it can do:

- Create a new incident (when something really bad happens)
- Update an existing incident (Found the cause, now working in a patch)
- Schedule a maintenance window (Weekly maintenance)
- Ends a maintenance or an incident (All operational again)

#### TODO

- [ ] Add tests (Need to patch requests and check request body)
- [x] Test locally using a real status page (personal page)


uilianries:
Closing as not planned to be a tool in Conan extensions.

PR #99, opened 2024-01-03
feature: Add token authentication to artifactory commands

### LINKED ISSUE #98: [feature] Add token-based authentication to art:server command
In our company, we have an Artifactory server with SSO. Somehow, it is imcompatible with the way `art:server add` tries to get a token because the URL it tries to access doesn't exist. However, I can manually create an identity token and I'd like a way to specify it, e.g.:

    conan art:server add my_art https://artifactory.company.com --user=fschoenm --token=<token>

There may be some corner cases but essentially, the subcommand to add a server should just use the specified token instead of trying to generate it.

danimtb:
Add token authentication for artifactory commands.

Closes #98

### DIFF
--- a/tests/test_artifactory_commands.py
+++ b/tests/test_artifactory_commands.py
@@ -328,3 +328,21 @@ def test_server_list_empty():
     out = run("conan art:server list")
 
     assert "No servers configured. Use `conan art:server add` command to add one." in out
+
+
+@pytest.mark.requires_credentials
+def test_add_server_token():
+    """
+    Test server add with token, list, remove commands
+    """
+
+    # Make sure artifactory repos are empty before starting the test
+    run("conan remove mypkg* -c -r extensions-stg")
+    run("conan remove mypkg* -c -r extensions-prod")
+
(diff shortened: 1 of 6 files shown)


tests/test_artifactory_commands.py
```diff
@@ -328,3 +328,21 @@ def test_server_list_empty():
...
+    """
+
+    # Make sure artifactory repos are empty before starting the test
+    run("conan remove mypkg* -c -r extensions-stg")
+    run("conan remove mypkg* -c -r extensions-prod")
+
+    server_url = os.getenv("ART_URL")
+    server_user = os.getenv("CONAN_LOGIN_USERNAME_EXTENSIONS_STG")
+
+    out_add = run(f'conan art:server add server1 {server_url} --user={server_user} --token={os.getenv("CONAN_PASSWORD_EXTENSIONS_STG")}')
```
danimtb, line 346: we need a token variable in CI for this test to work

czoido (approved):

Issue #109, opened 2024-01-12
[question] repository VS server VS url (vs remote)


h-2:
Why are there `repository`, `--server` and `--url`?  What's the difference, and why do I need multiple?

Also I get the following error:
```
  File "/home/hannes/.conan2/extensions/commands/art/cmd_build_info.py", line 167, in _get_remote_artifacts
    assert self._url and self._repository, "Missing information in the Conan local cache, " \
AssertionError: Missing information in the Conan local cache, please provide the --url and --repository arguments to retrieve the information from Artifactory.

ERROR: Missing information in the Conan local cache, please provide the --url and --repository arguments to retrieve the information from Artifactory.
```

Note that it says `--repository`, although "repository" is a positional argument and not specified via `--`.

It should also be noted that normal conan tools use the term "remote", accessible via `-r` and `--remote`. Why wasn't this used?

czoido:
Hi @h-2,
Thanks for reporting, actually we have to fix the ERROR information because it's misleading.
The difference between `repository`, `--server` and `--url` is:
- `repository` is the name of the repo in Artifactory where the artifacts are. It's important to note that is not the name of the conan remote. For example, if you have a local repo named `"staging"` where you have the artifacts to publish the build-info for, that is the `repository` argument.
- `--server` is the name of the server you manage with the `art:server` command. That's just the equivalent of setting the `--url` and `--password` arguments for the build-info commands that need credentials. If you supply both `--url` and `--password` and `--server` an exception should be raised to explain that only one of them should be used.
Did this help you?

PR #111, opened 2024-01-15
[fix] Fix and improve some help messages

danimtb:
closes https://github.com/conan-io/conan-extensions/issues/109


extensions/commands/art/cmd_build_info.py
```diff
@@ -173,7 +173,7 @@ def _get_local_artifacts():
         def _get_remote_artifacts(artifact):
             artifact_info = None
             assert self._url and self._repository, "Missing information in the Conan local cache, " \
-                                                   "please provide the --url and --repository arguments " \
+                                                   "please provide '--url' or '--server' arguments " \
```
czoido, line 176: I think we also want to change the assert above to not check the self._repository (although it should always be there...)

danimtb:
Replaced by #112

PR #117, opened 2024-02-19
Command to promote a pkglist form one Artifactory Pro repo to another on the same server using its API

RubenRBS:


extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,110 @@
...
+    if origin_remote == "Local Cache":
+        raise ConanException(f"Package list must come from the remote associated with {args.origin}, but found from local cache")
+
+    assert_server_or_url_user_password(args)
+
+    pkglist = multi_package_list[origin_remote]
+
+    for name_version, recipe in pkglist.recipes.items():
+        if "revisions" not in recipe:
+            ConanOutput().info(f"Recipe {name_version} does not have a revision, skipping")
```
RubenRBS, line 96: Maybe we want to upload everything if no rev?

extensions/commands/art/README.md
```diff
@@ -84,3 +84,35 @@ conan art:build-info promote mybuildname_aggregated 1 origin-artifactory-repo de
 ```
 
 Now, both release and debug binaries from that pacakge ae available in the destination repository with just one command.
+
+#### [conan art:promote_graph](cmd_promote_graph.py)
+
+```
+$ conan art:promote-graph -h
```
memsharded, line 91: Rename command, ``art:promote`` looks good to me

extensions/commands/art/README.md
```diff
@@ -84,3 +84,35 @@ conan art:build-info promote mybuildname_aggregated 1 origin-artifactory-repo de
...
+usage: conan promote-graph [-h] [-v [V]] [-cc CORE_CONF] [--server SERVER] [--url URL] [--user USER]
+                           [--password PASSWORD] [--token TOKEN]
+                           origin destination list
+
+Promote a pkglist file from an origin repository to a destination repository, without downloading the packages locally
+
+positional arguments:
+  origin                Artifactory origin repository.
```
memsharded, line 99: Name? URL?

extensions/commands/art/README.md
```diff
@@ -84,3 +84,35 @@ conan art:build-info promote mybuildname_aggregated 1 origin-artifactory-repo de
...
+  list                  Package list file to promote
+
+options:
+  -h, --help            show this help message and exit
+  -v [V]                Level of detail of the output. Valid options from less verbose to more verbose:
+                        -vquiet, -verror, -vwarning, -vnotice, -vstatus, -v or -vverbose, -vv or -vdebug,
+                        -vvv or -vtrace
+  -cc CORE_CONF, --core-conf CORE_CONF
+                        Global configuration for Conan
+  --server SERVER       Server name of the Artifactory to get the build info from
```
memsharded, line 110: buildInfo? probably this is outdated

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,110 @@
...
+from conans.model.recipe_ref import RecipeReference
+from conans.model.package_ref import PkgReference
+from conan.api.model import MultiPackagesList
+from conan.errors import ConanException
+
+from utils import api_request, assert_server_or_url_user_password
+from cmd_server import get_url_user_password
+
+
+def _get_export_path_from_rrev(rrev):
```
memsharded, line 16: We might want to do some ArtifactoryConanAPI common to all artifactory commands, these paths seems very reusable (not now, future)

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@conan_command(group="Artifactory")
+def promote_graph(conan_api: ConanAPI, parser, *args):
+    """
+    Promote a pkglist file from an origin repository to a destination repository, without downloading the packages locally
+    """
+    parser.add_argument("origin", help="Artifactory origin repository.")
+    parser.add_argument("destination", help="Artifactory destination repository.")
+    parser.add_argument("list", help="Package list file to promote")
```
memsharded, line 60: I always struggle with symlinks src->dst. Maybe ``conan art:promote mypkglist.json --from=repo1 --to=repo2``?

extensions/commands/art/cmd_promote_graph.py
```diff
@@ -0,0 +1,110 @@
...
+    listfile = os.path.realpath(args.list)
+    multi_package_list = MultiPackagesList.load(listfile)
+
+    remotes = list(multi_package_list.lists.keys())
+    if len(remotes) > 1:
+        raise ConanException(f"Expected every package to come from the same origin repository in {args.origin}")
+
+    origin_remote = remotes[0]
+    if origin_remote == "Local Cache":
+        raise ConanException(f"Package list must come from the remote associated with {args.origin}, but found from local cache")
```
memsharded, line 88: It could accept an optional ``--remote`` name to select packages from, in case a packagelist with multiple remotes?

extensions/commands/art/cmd_promote_graph.py
memsharded: I'd say that promoting many revisions seems mostly an anti-pattern, you could be promoting a lot of garbage.

danimtb (changes requested):
I left some comments in the review, but in general I have some doubts regarding the remote and origin/destination params.

- Why do you need to specify remote or origin repo if you already have it in the pkglist file?
- Keep in mind that conan remote name can be different from Artifactory repository name, so it should be clear to the user what should be used.

extensions/commands/art/cmd_promote.py
```diff
@@ -0,0 +1,122 @@
...
+
+@conan_command(group="Artifactory")
+def promote(conan_api: ConanAPI, parser, *args):
+    """
+    Promote a pkglist file from an origin Artifactory repository to a destination repository, without downloading the packages locally
+    """
+
+    parser.add_argument("list", help="Package list file to promote")
+    parser.add_argument("--from", help="Artifactory origin repository name", required=True, dest="origin")
+    parser.add_argument("--to", help="Artifactory destination repository name", required=True, dest="destination")
```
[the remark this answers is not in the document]
RubenRBS: The pkglist only represents the name of the origin given in `conan remote add <name>`, which as you pointed out, might very well differ from the artifactory name, so these parameters are meant to solve that, by letting you specify the names of the actual remote.

Maybe a better UX would be to have Conan find the proper name from the API, but I don't know if that's possible. Any insight?

extensions/commands/art/cmd_promote.py
```diff
@@ -0,0 +1,122 @@
...
+
+    parser.add_argument("--remote", help="Remote name to use for the origin repositories", default=None)
```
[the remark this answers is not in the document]
RubenRBS: Reads much better!

extensions/commands/art/cmd_promote.py
```diff
@@ -0,0 +1,122 @@
...
+        ConanOutput().warning("Destination already exists, skipping")
+    except ConanException:
+        _request(url, user, password, "post", f"api/copy/{origin}/{path}?to=/{destination}/{path}&suppressLayouts=0")
+        ConanOutput().success("Promoted file")
+
+
+@conan_command(group="Artifactory")
+def promote(conan_api: ConanAPI, parser, *args):
+    """
+    Promote a pkglist file from an origin Artifactory repository to a destination repository, without downloading the packages locally
```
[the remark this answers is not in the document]
RubenRBS: This can also be used to promote just the recipes if the pkglist contains no packages, but it's true that something a bit more clear is needed. Any suggestions that might let the user know that it will promote everything?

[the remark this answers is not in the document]
danimtb:
```suggestion
    Promote Conan recipes and packages in a pkglist file from an origin Artifactory repository to a destination repository, without downloading the packages locally
```
Maybe like so?

extensions/commands/art/cmd_promote.py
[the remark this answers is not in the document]
danimtb: That was discussed previously with @czoido but it is very fragile to rely on the conan remote's URL to get the artifactory repo name, as the URL of the repo can be behind another domain, like center.conan.io.

IMO, if the `--remote` argument is only needed to "select" which packages should be promoted, I'd drop this argument and assume all the pkglist packages are in the origin repo and error if they aren't. For me this behavior seems more straightforward.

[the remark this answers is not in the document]
RubenRBS: But the issue remains, we have no idea to know the name of the remote in Artifactory without explicitly being told, but maybe I'm missing something? :)

extensions/commands/art/README.md
```diff
@@ -84,3 +84,41 @@ conan art:build-info promote mybuildname_aggregated 1 origin-artifactory-repo de
 ```
 
 Now, both release and debug binaries from that pacakge ae available in the destination repository with just one command.
+
+#### [conan art:promote](cmd_promote.py)
+
+```
+$ conan art:promote -h
```
danimtb, line 91: Updating the help message is the only thing left 💪

RubenRBS: Good catch!

danimtb (approved):

PR #123, opened 2024-05-08
Improvements to the PDB hook readme

juansblanco:
Added section explaining how to install hook to readme and fixed some typos.


extensions/hooks/README.md
```diff
@@ -3,14 +3,28 @@ This hook copies the PDBs from their original location in the build folder to th
 This is required for debugging libraries with Visual Studio when the original source files aren't present.
 For more information on how to debug using the hook check the [documentation](https://docs.conan.io/2/examples/dev_flow/debug/debugging_visual.html)
 
+### Installing the hook
+
+The hook can be installed with a `conan config install` with all the conan extensions, but it is recommended to install
+just the hook with:
```
czoido, line 9: I would summarize a bit:

```suggestion
To install just the hooks from this repository, without the other extensions, use the command:
```

extensions/hooks/README.md
```diff
@@ -3,14 +3,28 @@ This hook copies the PDBs from their original location in the build folder to th
...
+
+`conan config install https://github.com/conan-io/conan-extensions.git -sf=extensions/hooks -tf=extensions/hooks`
+
+The hook is made so it won't run by default, as it can increase the package size significantly. As explained in the
+[hooks documentation](https://docs.conan.io/2/reference/extensions/hooks.html), we need to change the name of our hook to start with `hook_`.
+To locate the path where the hook was placed, run the command `conan config home` to find
+your local cache path and go to the `extensions/hooks` folder to rename the `_hook_copy_pdbs_to_package.py` file.
+Be aware that this hook will run everytime a `package()` method is run, to disable the hook just rename the hook back
+to start with `_hook_`.
```
czoido, line 18: Bear in mind that this is a README common for all the hooks

czoido: You can start with something like The hooks are named so they won't run by default... then the part of the size is exclusive from the PDB's hook, but anyway we don't want that the hooks run by default for whatever the reason could be.

extensions/hooks/README.md
```diff
@@ -1,3 +1,16 @@
+## Installing hooks
+
+To install just the hooks from this repository, without the other extensions, use the command:
+
+`conan config install https://github.com/conan-io/conan-extensions.git -sf=extensions/hooks -tf=extensions/hooks`
+
+Some hooks are made so they won't be run by default, so we need to change the name of the hook
```
czoido, line 7:
```suggestion
The hooks are named so they won't be run by default, so we need to change the name of the hook
```

czoido (approved):

PR #125, opened 2024-05-14
Support for air-gapped envs for backup sources

RubenRBS:


extensions/commands/x/cmd_backups.py
```diff
@@ -0,0 +1,41 @@
...
+        checksum = sha256(f.read())
+
+    ConanOutput().info(f"Calculated checksum: {checksum}")
+    new_path = os.path.join(os.path.dirname(args.path), checksum)
+    json_path = new_path + ".json"
+    shutil.move(args.path, new_path)
+    try:
+        with open(json_path, "w") as f:
+            json.dump({"references": {args.reference: [args.url] if args.url else []},
+                       "timestamp": timestamp_now()}, f)
```
uilianries, line 35: Why not using conan.tools.files.save instead? Just asking in case there is some advantage.

extensions/commands/x/cmd_backups.py
```diff
@@ -0,0 +1,41 @@
+import os
+import shutil
+import json
+
+from conan.api.output import ConanOutput
+from conan.cli.command import conan_command
+from conans.errors import ConanException
```
uilianries, line 7:
```suggestion
from conan.errors import ConanException
```

danimtb:
I think a bit of context is needed in the description of the PR and a test with an example os usage will be nice as well if possible (maybe it is difficult to mock). Otherwise, an example for documentation

extensions/commands/x/cmd_backups.py
RubenRBS: @uilianries sorry for the delay! `save()` needs a conanfile as the first argument, so we can't use it here :)

extensions/commands/x/cmd_backups.py
```diff
@@ -0,0 +1,41 @@
...
+    ConanOutput().info(f"Calculated checksum: {checksum}")
+    new_path = os.path.join(os.path.dirname(args.path), checksum)
+    json_path = new_path + ".json"
+    shutil.move(args.path, new_path)
+    try:
+        with open(json_path, "w") as f:
+            json.dump({"references": {args.reference: [args.url] if args.url else []},
+                       "timestamp": timestamp_now()}, f)
+
+        conan_api.upload.upload_backup_sources([new_path, json_path])
```
memsharded, line 37: This will silently return if ``url = config.get("core.sources:upload_url", check_type=str)`` is not defined. Most likely we want to check this and error?

extensions/commands/x/cmd_backups.py
```diff
@@ -0,0 +1,41 @@
...
+from conans.util.sha import sha256
+
+
+@conan_command(group="Backup sources handling")
+def backups(conan_api, parser, *args):
+    """
+    This command will upload the backup sources of a package to the server.
+    """
+    parser.add_argument("reference", help="Reference of the package to backup")
+    parser.add_argument("path", help="Path to source files for this reference")
```
memsharded, line 18: Provide the possibility to download it too from a URL?

extensions/commands/x/cmd_backups.py
```diff
@@ -0,0 +1,41 @@
...
+    parser.add_argument("--url", help="Original URL of the file", default=None)
+    args = parser.parse_args(*args)
+
+    if not os.path.isfile(args.path):
+        raise ConanException(f"File not found: {args.path}")
+
+    with open(args.path, "rb") as f:
```
memsharded, line 25: I'd say to include the checksum input and verification as well? Not just read the local one, could be tampered?

memsharded (commented):
I think I was reading this wrong. This is intended to be done after the user has manually downloaded and verified the file checksum, isn't it? 

Otherwise, do we want to provide the possibility to download it too from a URL?
Though maybe if we want this, it would be 2 commands:
- 1 command download + verify checksum (this wouldn't be necessary for Linux/OSX users, though Windows bare users might benefit? Probably not worth
- 1 command for upload

PR #134, opened 2024-06-07
Add build_url support

schwaerz:
As requested in #133.

Currently I see two possible issues with this PR:
1. No automated tests. I could not get the tests running locally. Maybe you can help me out a bit.
2. The current implementation would be a breaking change as it requires you to set the `build_url` as an additional mandatory argument to `art:build_info create`. Probably it should be an optional argument? I was not sure - but wanted to implement it in the same fashion like the build number and name was implemented.

PR #137, opened 2024-06-25
Fix BuildInfo arguments

czoido:
Fixes: https://github.com/conan-io/conan-extensions/pull/134#issuecomment-2186996748


RubenRBS (approved):

markferry:
Thanks for the quick turnaround!

In the meantime I had forked and tagged for my purposes. I shouldn't have left a moving target in a production CI system anyway.

Would it be premature to suggest upstream tagging here? (Just a simple CalVer perhaps)

So everyone can:

`conan config install -t git -a '-b some-tag --depth=1' https://github.com/conan-io/conan-extensions`

PR #142, opened 2024-07-12
[fix] Fix hook copy pdbs for recipes with win_bash=True

### LINKED ISSUE #135: hook_copy_pdbs_to_package.py fails for ffmpeg on Windows MSVC with MSYS
I assume it is failing because `conanfile.run()` is executing in the msys environment (ie, what ffmpeg is building inside)

This is what it printed out:

```
INSTALL m/cocache4/cache/b/ffmpebf64c855994b7/b/src/libavutil/tea.h
INSTALL m/cocache4/cache/b/ffmpebf64c855994b7/b/src/libavutil/tx.h
INSTALL m/cocache4/cache/b/ffmpebf64c855994b7/b/src/libavutil/film_grain_params.h
INSTALL m/cocache4/cache/b/ffmpebf64c855994b7/b/src/libavutil/video_hint.h
INSTALL libavutil/avconfig.h
INSTALL libavutil/ffversion.h

ffmpeg/6.1: [HOOK - hook_copy_pdbs_to_package.py] post_package(): PDBs post package hook running
ffmpeg/6.1: [HOOK - hook_copy_pdbs_to_package.py] post_package(): RUN: "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -find "**\dumpbin.exe" -format json
mkdir: cannot create directory ‘/dev/shm’: Read-only file system
mkdir: cannot create directory ‘/dev/mqueue’: Read-only file system
/usr/bin/bash: line 1: fg: no job control

ERROR: [HOOK - hook_copy_pdbs_to_package.py] post_package(): Failed to locate dumpbin.exe which is needed to locate the PDBs and copy them to package folder.
```

danimtb:
closes #135 

There were some errors in this hook when it was executed in recipes that were running in a bash shell in windows (recipe attribute `win_bash=True`). Mainly because of the variable shell expansion, path issues and CL flag formatting.

Also included another fix for PDBs that are already copied (by cmake install command) to the package folder.

### DIFF
--- a/extensions/hooks/_hook_copy_pdbs_to_package.py
+++ b/extensions/hooks/_hook_copy_pdbs_to_package.py
@@ -5,6 +5,7 @@ import os
 import re
 from io import StringIO
 from conan.errors import ConanException
+from conans.util.env import get_env
 
 
 def post_package(conanfile):
@@ -18,23 +19,32 @@ def post_package(conanfile):
     # Find dumpbin path
     output = StringIO()
     try:
+        program_files = get_env("ProgramFiles(x86)") or get_env("ProgramFiles")
         conanfile.run(
-            r'"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -find "**\dumpbin.exe" -format json',
+            rf'"{program_files}\Microsoft Visual Studio\Installer\vswhere.exe" -find "**\dumpbin.exe" -format json',
             stdout=output, scope="")
+        match = re.search(r'\[(.*?)\]', str(output.getvalue()), re.DOTALL)
+        dumpbin_path = json.loads(f'[{match.group(1)}]')[0]
     except ConanException:
         raise ConanException(
             "Failed to locate dumpbin.exe which is needed to locate the PDBs and copy them to package folder.")
-    dumpbin_path = json.loads(str(output.getvalue()))[0]
 
     for dll_path in package_dll:
         # Use dumpbin to get the pdb path from each dll
         dumpbin_output = StringIO()
-        conanfile.run(rf'"{dumpbin_path}" /PDBPATH {dll_path}', stdout=dumpbin_output)
+        pdbpath_flag = "-PDBPATH" if conanfile.win_bash else "/PDBPATH"
+        conanfile.run(rf'"{dumpbin_path}" {pdbpath_flag} "{dll_path}"', stdout=dumpbin_output)
         dumpbin = str(dumpbin_output.getvalue())
         pdb_path = re.search(r"[“'\"].*\.pdb[”'\"]", dumpbin)
         if pdb_path:
             pdb_path = pdb_path.group()[1:-1]
-            # Copy the corresponding pdb file from the build to the package folder
-            conanfile.output.info(
-                f"copying {os.path.basename(pdb_path)} from {os.path.dirname(pdb_path)} to {os.path.dirname(dll_path)}")
-            copy(conanfile, os.path.basename(pdb_path), os.path.dirname(pdb_path), os.path.dirname(dll_path))
+            pdb_file = os.path.basename(pdb_path)
+            src_path = os.path.dirname(pdb_path)
+            dst_path = os.path.dirname(dll_path)
+            if src_path != dst_path:  # if pdb is not allready in the package folder, then copy
+                # Copy the corresponding pdb file from the build to the package folder
+                conanfile.output.info(
+                    f"copying {pdb_file} from {src_path} to {dst_path}")
+                copy(conanfile, os.path.basename(pdb_path), os.path.dirname(pdb_path), os.path.dirname(dll_path))
+            else:
+                conanfile.output.info(f"PDB file {pdb_file} already in destination folder {dst_path}, skipping copy")


extensions/hooks/_hook_copy_pdbs_to_package.py
```diff
@@ -18,23 +19,32 @@ def post_package(conanfile):
...
+            pdb_file = os.path.basename(pdb_path)
+            src_path = os.path.dirname(pdb_path)
+            dst_path = os.path.dirname(dll_path)
+            if src_path != dst_path:  # if pdb is not allready in the package folder, then copy
+                # Copy the corresponding pdb file from the build to the package folder
+                conanfile.output.info(
+                    f"copying {pdb_file} from {src_path} to {dst_path}")
+                copy(conanfile, os.path.basename(pdb_path), os.path.dirname(pdb_path), os.path.dirname(dll_path))
+            else:
+                conanfile.output.info(f"PDB file {pdb_file} already in destination folder {dst_path}, skipping copy")
```
danimtb, line 50: as example to add this change, I bump into this:

```
freetype/2.13.2: [HOOK - hook_copy_pdbs_to_package.py] post_package(): copying freetype-6.pdb from C:\Users\danielm\.conan2\p\b\freet9a8e0c06e6918\b\build-debug to C:\Users\danielm\.conan2\p\b\freet9a8e0c06e6918\p\bin
freetype/2.13.2: [HOOK - hook_copy_pdbs_to_package.py] post_package(): RUN: "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\bin\Hostx64\x64\dumpbin.exe" /PDBPATH "C:\Users\danielm\.conan2\p\b\freet9a8e0c06e6918\p\bin\freetyped.dll"
freetype/2.13.2: [HOOK - hook_copy_pdbs_to_package.py] post_package(): copying freetype-6.pdb from C:\Users\danielm\.conan2\p\b\freet9a8e0c06e6918\p\bin to C:\Users\danielm\.conan2\p\b\freet9a8e0c06e6918\p\bin
ERROR: [HOOK - hook_copy_pdbs_to_package.py] post_package(): copy() 'src' and 'dst' arguments must have different values
```

extensions/hooks/_hook_copy_pdbs_to_package.py
```diff
@@ -5,6 +5,7 @@
 import re
 from io import StringIO
 from conan.errors import ConanException
+from conans.util.env import get_env
```
danimtb, line 8: Not sure of adding this import from `conans`. Maybe it should be migrated to `conan` module first?

extensions/hooks/_hook_copy_pdbs_to_package.py
```diff
@@ -18,23 +19,32 @@ def post_package(conanfile):
     # Find dumpbin path
     output = StringIO()
     try:
+        program_files = get_env("ProgramFiles(x86)") or get_env("ProgramFiles")
         conanfile.run(
-            r'"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -find "**\dumpbin.exe" -format json',
+            rf'"{program_files}\Microsoft Visual Studio\Installer\vswhere.exe" -find "**\dumpbin.exe" -format json',
             stdout=output, scope="")
+        match = re.search(r'\[(.*?)\]', str(output.getvalue()), re.DOTALL)
```
danimtb, line 26: I found some lines in the json oputput that where not json (I believe they where some warnings or something). So I had to find the json list in the output to process it first 🤷‍♂️

extensions/hooks/_hook_copy_pdbs_to_package.py
```diff
@@ -18,23 +19,32 @@ def post_package(conanfile):
     # Find dumpbin path
     output = StringIO()
     try:
+        program_files = get_env("ProgramFiles(x86)") or get_env("ProgramFiles")
```
jcar87, line 22: I'd say we can use `os.environ.get` in this case, rather than the legacy `get_env`

Or alternatively the string for the command can use `%VAR%` vs `$VAR` conditional on winbash, if we can evaluate it.

I'd also add a comment that `get_env("ProgramFiles")` is a fallback for 32-bit windows

extensions/hooks/_hook_copy_pdbs_to_package.py
jcar87: mmm do you have an example of the output?

extensions/hooks/_hook_copy_pdbs_to_package.py
danimtb: I had some issues with variable expansion and the special characters in bash, and after a few tries I ipted for this approach. Will use `os.environ.get`, thanks!

extensions/hooks/_hook_copy_pdbs_to_package.py
danimtb: I do not have it exactly but it was something like:

```
[vcvarsall.bat] Environment initialized for: 'x64'
mkdir: cannot create directory ‘/dev/shm’: Read-only file system

Creating /dev/shm directory failed.
POSIX semaphores and POSIX shared memory will not work

mkdir: cannot create directory ‘/dev/mqueue’: Read-only file system

Creating /dev/mqueue directory failed.
POSIX message queues will not work

[ "this was the actual list of devenv paths",
  "here as well",
  "and here"
]
```
maybe a missconfiguration of my environment?

jcar87: I see, interesting
this is a consequence of of `self.run` running with the bash enabled (so we get all the other output)
mmm
i wonder if we can disable running via the shell (win bash) for a specific call

danimtb: I see no parameter in the ``run`` function itself to override the recipe attribute https://github.com/conan-io/conan/blob/1d8fb4f7289d3676f8165e514d799095a10a3cb2/conans/model/conan_file.py#L340

We could change the attribute just for the hook run calls and revert its value at the end, but I am unsure about that approach and its consequences.

AbrilRBS (approved):

jcar87 (approved):

PR #150, opened 2024-08-21
Add Ament generator for ROS2 colcon builds

danimtb:
Initial commit for the Ament generator.

There is still much to polish and improve, including files generated (not all might be needed) and output.


danimtb:
Pinging you here @czoido @AbrilRBS so you are aware that this is going to be merged and added to this repo to keep maturing the generator. Thanks!

PR #151, opened 2024-08-22
Fix: Fix CI for artifactory tests colliding in parallel run

danimtb:
There is currently an issue in the CI for tests that require interacting with Artifactory, as it happened at #150.
Those tests are run in parallel and sometimes they interfere, causing the CI to fail.

This PRs makes those tests run sequentially (unfortunately I did not find a more elegant way to achieve that 😓)

### DIFF
--- a/.github/workflows/test_conan_extensions.yml
+++ b/.github/workflows/test_conan_extensions.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install -U pip
-        pip install pytest "cyclonedx-python-lib>=5.0.0,<6"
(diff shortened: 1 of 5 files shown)


AbrilRBS (commented):
Oh, nice catch!
But as far as I can tell, isn't this a plugin that would need to be installed first? https://pytest-dev.github.io/pytest-order/stable/intro.html

danimtb:
> Oh, nice catch! But as far as I can tell, isn't this a plugin that would need to be installed first? https://pytest-dev.github.io/pytest-order/stable/intro.html

Yes, you are right. I also checked it beforehand but it seemed to work in my machine and also in the CI. Anyway, I will add it to the pip install call and try to remove some of the warnings in the output as well. Thanks!

PR #155, opened 2024-09-11
Fix empty buildinfo when using conan export-package

getrostt:
When building a conan package using `conan export-pkg`, the resulting build-info when running `conan art:build-info create` is empty. This is caused by an unconditional removal of the 1st graph node in the resulting json file.

This fix only removes the 1st node when it is a ref of type `conanfile`.

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -389,7 +389,8 @@ def build_info_create(conan_api: ConanAPI, parser, subparser, *args):
     data = load_json(args.json)
 
     # remove the 'conanfile' node
-    data["graph"]["nodes"].pop("0")
+    if data["graph"]["nodes"]["0"]["ref"] == "conanfile":
+        data["graph"]["nodes"].pop("0")
     bi = _BuildInfo(data, args.build_name, args.build_number, args.repository,
                     build_url=args.build_url,
                     with_dependencies=args.with_dependencies,
--- a/tests/test_artifactory_commands.py
+++ b/tests/test_artifactory_commands.py
@@ -141,6 +141,28 @@ def test_build_info_create_with_build_url():
 
 
 @pytest.mark.requires_credentials
+def test_build_info_create_exppkg():
+    build_name = "mybuildinfo"
+    build_number = "1"
+
+    # Configure Artifactory server and credentials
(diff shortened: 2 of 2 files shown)


extensions/commands/art/cmd_build_info.py
```diff
@@ -389,7 +389,8 @@ def build_info_create(conan_api: ConanAPI, parser, subparser, *args):
     data = load_json(args.json)
 
     # remove the 'conanfile' node
-    data["graph"]["nodes"].pop("0")
+    if data["graph"]["nodes"]["0"]["ref"] == "conanfile":
```
memsharded, line 392: I am concern about cases where the consumer might have defined a ``name`` and ``version``, so the ``ref`` might not be "conanfile", but the actual ``name/version``. 

Maybe it would be necessary to add a test for that scenario.

memsharded (commented):
Thanks very much for your contribution @getrostt 

Adding @czoido so he might be able to help checking the potential issue (maybe the way to avoid it is checking the existing of the ``path`` field?

extensions/commands/art/cmd_build_info.py
getrostt: I tried to replay the scenario on the commandline using the following commands:
```bash
conan new cmake_lib -d name=mypkg -d version=1.0 --force
conan create . --format json -tf='' -s build_type=Release > create_release.json
```

That results in the following `create_release.json` (I only pasted the 1st lines):
```json
{
    "graph": {
        "nodes": {
            "0": {
                "ref": "conanfile",
                "id": "0",
```

Is this the behavior you were thinking of for a test?

extensions/commands/art/cmd_build_info.py
```diff
@@ -389,7 +389,8 @@ def build_info_create(conan_api: ConanAPI, parser, subparser, *args):
...
+        data["graph"]["nodes"].pop("0")
```
czoido, line 393:
```suggestion
    if not data["graph"]["nodes"]["0"]["prev"]:
        data["graph"]["nodes"].pop("0")
```
I think this should work well in all cases

czoido:
Hi @getrostt,
Thanks a lot for the contribution. I proposed some changes there to consider all cases. I'll have to open a separate PR for this because we can only run tests marked as requires_credentials if they run from a branch from the repo not from forks. I'll cherry-pick your changes so you still appear as contributor.

PR #157, opened 2024-09-12
Fix empty buildinfo when using conan export-package

czoido:
Copy of this one: https://github.com/conan-io/conan-extensions/pull/155
As commented by @getrostt in the original PR:

>When building a conan package using `conan export-pkg`, the resulting build-info when running `conan art:build-info >create` is empty. This is caused by an unconditional removal of the 1st graph node in the resulting json file.

To run tests with credentials

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -389,7 +389,7 @@ def build_info_create(conan_api: ConanAPI, parser, subparser, *args):
     data = load_json(args.json)
 
     # remove the 'conanfile' node
-    if data["graph"]["nodes"]["0"]["ref"] == "conanfile":
+    if not data["graph"]["nodes"]["0"]["prev"]:
         data["graph"]["nodes"].pop("0")
     bi = _BuildInfo(data, args.build_name, args.build_number, args.repository,
                     build_url=args.build_url,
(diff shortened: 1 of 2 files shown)


extensions/commands/art/cmd_build_info.py
```diff
@@ -389,7 +389,8 @@ def build_info_create(conan_api: ConanAPI, parser, subparser, *args):
     data = load_json(args.json)
 
     # remove the 'conanfile' node
-    data["graph"]["nodes"].pop("0")
+    if not data["graph"]["nodes"]["0"]["prev"]:
```
danimtb, line 392: what is the meaning of "prev" here? Like _previous_ node in the graph? _Previous_ dependency? Maybe it could be useful to update the comment in the upper line with a small clarification

czoido: Yes, I think it would be good to add a comment there, `prev` it's the field that stores the package revision in the graph json output, no `prev` means that it has no binaries

czoido, line 392:
```suggestion
    # remove only if prev=null that means that the node has no binaries
    if not data["graph"]["nodes"]["0"]["prev"]:
```

czoido: Added a comment there

danimtb (approved):

memsharded (commented):
Maybe this was superseded by recent fixes in ``conan export-pkg`` build-info like https://github.com/conan-io/conan-extensions/pull/201?

czoido:
> Maybe this was superseded by recent fixes in `conan export-pkg` build-info like #201?

Yes, I think we can close this one.

PR #159, opened 2024-09-24
Add a --no-build-requires command line option to the sbom:cyclonedx extension command

### LINKED ISSUE #158: [improvement] SBOM extension command sbom:cyclonedx - add new --no-build-requires command line argument
I would like propose some code to add a new command line option to the SBOM extension command sbom:cyclonedx to allow to filter the graph node based on dependency context (host and/or build).
While Ken Thompson's thesis about injecting a virus into a compiler is interesting, we'd like to ignore that and generate an SBOM that doesn't include the build_requires :-)
The change to do that is simple but I'd like first to have guidance about how the command line option should be presented. i.e. how to name it, what syntax should it have etc.
Also, in the source code of the extension, there's a FIXME note about some --build-require option, can someone explain what this is about ?

johan-boule:
sbom:cyclonedx --no-build-requires will omit build requirements from the generated SBOM, i.e. keeping only dependencies that are in the host context.

closes #158

### DIFF
--- a/extensions/commands/sbom/cmd_cyclonedx.py
+++ b/extensions/commands/sbom/cmd_cyclonedx.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Iterable, List, Optional, Set, Tuple, Union
 
 from conan.api.conan_api import ConanAPI
 from conan.api.output import cli_out_write
+from conan.api.subapi.graph import CONTEXT_BUILD
 from conan.cli.args import common_graph_args, validate_common_graph_args
 from conan.cli.command import conan_command
 from conan.errors import ConanException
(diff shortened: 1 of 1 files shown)


extensions/commands/sbom/cmd_cyclonedx.py
```diff
@@ -151,12 +154,16 @@ def me_as_tool() -> Tool:
                                                          remotes, args.update)
     # endregion COPY
 
-    components = {node: create_component(node) for node in deps_graph.nodes}
+    def filter_context(node): return not args.no_build_requires or node.context != CONTEXT_BUILD
```
memsharded, line 157:
```suggestion
    def filter_context(n): return not args.no_build_requires or n.context != CONTEXT_BUILD
```
It is generally better to avoid using arguments named as existing variables in scope

memsharded (commented):
Thanks for your contribution

PR #175, opened 2025-02-12
Fix: Zip Deployer crashes when subfolder exist

Leanny:
Right now, the zip deployer crashes with an exception when a subfolder exists inside bin or lib. This can easily be checked with

```
conan install --requires=boost/1.85.0 --deployer=runtime_zip_deploy`
```

The change add sone unittest with a minimal project that creates a folder inside `bin` to ensure that it does not crash like the current version does.

### DIFF
--- a/extensions/deployers/runtime_zip_deploy.py
+++ b/extensions/deployers/runtime_zip_deploy.py
@@ -3,6 +3,23 @@ import shutil
 import zipfile
 
 
+def _get_files_from_folder(search_dir, output_folder):
+    # recusively copy all .dlls and .exes from the search_dir to the output_folder
+    if not os.path.isdir(search_dir):
+        return []
+    files = []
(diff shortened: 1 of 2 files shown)


extensions/deployers/runtime_zip_deploy.py
```diff
@@ -3,6 +3,23 @@
 import zipfile
 
 
+def _get_files_from_folder(search_dir, output_folder):
+    # recusively copy all .dlls and .exes from the search_dir to the output_folder
+    if not os.path.isdir(search_dir):
+        return []
+    files = []
+    for f in os.listdir(search_dir):
```
AbrilRBS, line 11: I wonder if using os.walkdir is a better approach here, but I like the current minimal diff nevertheless

danimtb (approved):

AbrilRBS (approved):

Issue #191, opened 2025-04-16
Artifactory --out-file Creates Empty File


BlackAngelTiger:
Hello Maintainers,

I wanted to use the `--out-file` parameter which was mentioned in the `extensions/commands/art/readme_build_info.md`, but I am getting a empty output file. I want to use the output file, since this is easier to use in our python scripts than the redirection of the stdout.

The build-info JSON output of the command was still printed to the console (replaced the `modules` section with `...` to keep it shorter):

```Console
{
    "version": "1.0.1",
    "name": "local-test-build",
    "number": "local",
    "agent": {},
    "started": "2025-04-14T08:05:33.585+0200",
    "buildAgent": {
        "name": "conan",
        "version": "2.15.0"
    },
   ...
}
Formatted output saved to './build/art-build-info.json'
```

The command used for the generation was:
`conan art:build-info create ./build/create-out.json local-test-build local testing --server pkg --out-file ./build/art-build-info.json`

The `create-out.json` contained valid information, so this does not cause the empty output file.

I am using Conan `2.15.0`, and the `3a7772f3e255479628a05705c5ff1f4b8bbd83a7` commit (main of 27.03.2025) of this repository.

What am I doing wrong or is this a Bug?

czoido:
Hi @BlackAngelTiger,

Thank you very much for reporting this issue.
You're absolutely right, this appears to be a bug in the output of the `conan art:build-info` command, which currently doesn't support the `--out-file` argument. This argument was added relatively recently, and the command wasn't updated to handle it properly.
I'll work on a fix and hopefully get it resolved soon.

czoido:
https://github.com/conan-io/conan-extensions/pull/192 should fix it

BlackAngelTiger:
Thank you for the super quick response.  I have tested it with the latest main and it works like a charm. Thank you very match.

memsharded:
Thanks to you for reporting!

PR #192, opened 2025-04-16
Enable out-file for build info commands

czoido:
For --out-file to work you have to use the formatters for the output, otherwise Conan can't capture the stdout to write the file.
Closes: https://github.com/conan-io/conan-extensions/issues/191

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -352,6 +352,11 @@ def _add_default_arguments(subparser, is_bi_create=False, is_bi_create_bundle=Fa
     return subparser
 
 
+def text_formatter(info):
+    if info:
+        cli_out_write(info)
+
+
 @conan_command(group="Artifactory")
 def build_info(conan_api: ConanAPI, parser, *args):
     """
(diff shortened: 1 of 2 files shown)


extensions/commands/art/cmd_build_info.py
```diff
@@ -352,14 +352,19 @@ def _add_default_arguments(subparser, is_bi_create=False, is_bi_create_bundle=Fa
...
+
 @conan_command(group="Artifactory")
 def build_info(conan_api: ConanAPI, parser, *args):
     """
     Manages JFrog Build Info (https://www.buildinfo.org/)
     """
 
 
-@conan_subcommand()
+@conan_subcommand(formatters={"text": text_formatter})
```
memsharded, line 367: Being the build-info a json file this is a bit confusing, but I understand this is the default formatter, so 👍

extensions/commands/art/cmd_build_info.py
```diff
@@ -437,12 +442,12 @@ def build_info_upload(conan_api: ConanAPI, parser, subparser, *args):
         request_url = f"{request_url}?project={args.project}"
     response = api_request("put", request_url, user, password, json_data=json.dumps(build_info_json))
     if response:
-        cli_out_write(response)
+        return response
     else:
-        cli_out_write("Build info uploaded successfully.")
+        return "Build info uploaded successfully."
```
memsharded, line 447: This is weird as a result, it seems more like a ``ConanOutput()`` informational message

extensions/commands/art/cmd_build_info.py
czoido: yes, I agree, It's a bit confusing

PR #193, opened 2025-04-22
[New Command] to generate diff between upstream versions and view it in HTML

ErniGH:


extensions/commands/diff/cmd_check_diff.py
```diff
@@ -0,0 +1,176 @@
...
+    parser.add_argument("v1", help="Old version")
+    parser.add_argument("v2", help="New version")
+    parser.add_argument("-s", "--split_diff", action="store_true")
+    args = parser.parse_args(*args)
+
+    cwd = os.getcwd()
+    path = conan_api.local.get_conanfile_path(args.path, cwd, py=True)
+    enabled_remotes = conan_api.remotes.list()
+
+    src_dir = os.path.join(path, "src") #TODO can break if the src_folder is not defined as "src"
```
AbrilRBS, line 28: I know we talked about using the sources if they are available from the Conan cache, but, if they are not:

Make this 2 temp folders, and call `source()` on each of them, then diff both folders directly using the `diff` command or whatever python library is available.

This will let you skip creating a git repo and the flow won't break if the source folder is not named src

AbrilRBS (commented):
Some comments

extensions/commands/diff/cmd_check_diff.py
AbrilRBS, line 28:
```suggestion
    src_dir = os.path.join(os.path.dirname(path), "src") #TODO can break if the src_folder is not defined as "src"
```
So that you don't try to create a subfolder of the conanfile.py file

AbrilRBS:
Closing in favour of making this a core client function in https://github.com/conan-io/conan/pull/18247

PR #197, opened 2025-06-17
Command for support of macOS/iOS universal binaries

gmeeker:
This is an attempt to build on the CMake support for '-s arch=armv8|x86_64' by pretending that all generates can produce universal binaries. When called to package a recipes as universal binaries, it bypasses the build() / package() and runs lipo with the single architecture binaries instead.

This currently duplicate the code of the install command and calls a lot of non-public API. It also executes install with direct_deploy for each architecture. This could be more efficient, both by avoiding the copy and moving this directly into Conan's dependency graph. However, it seems to work.

Note that this currently builds single architectures even for CMake recipes. There is no guarantee that all CMake recipes support this (for example they may detect the architecture to use ARM or x86 assembly). Recipes would probably need a way to opt-in for universal CMake builds.

### DIFF
--- a/extensions/commands/install-universal/cmd_install_universal.py
+++ b/extensions/commands/install-universal/cmd_install_universal.py
@@ -1,0 +1,248 @@
+import json
+import os
+import shutil
+from subprocess import run
+import sys
+import tempfile
+
+from conan.api.conan_api import ConanAPI
+from conan.api.output import ConanOutput
+from conan.cli import make_abs_path
+from conan.cli.args import common_graph_args, validate_common_graph_args
+from conan.cli.command import conan_command
+from conan.cli.formatters.graph import format_graph_json
+from conan.cli.printers import print_profiles
+from conan.cli.printers.graph import print_graph_packages, print_graph_basic
+
+
+@conan_command(group="Consumer", formatters={"json": format_graph_json})
+def install_universal(conan_api: ConanAPI, parser, *args):
+    """
+    Install universal packages from the requirements specified in a recipe (conanfile.py or conanfile.txt).
+
+    It can also be used to install packages without a conanfile, using the
+    --requires and --tool-requires arguments.
+
+    If any requirement is not found in the local cache, it will iterate the remotes
+    looking for it. When the full dependency graph is computed, and all dependencies
+    recipes have been found, it will look for binary packages matching the current settings.
+    If no binary package is found for some or several dependencies, it will error,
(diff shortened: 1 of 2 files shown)


extensions/commands/install-universal/cmd_install_universal.py
```diff
@@ -0,0 +1,248 @@
...
+        for arch in archs:
+            more_args = ''
+            if args.profile_build:
+                more_args += f' -pr:b {args.profile_build[0]}'
+            if args.profile_host:
+                more_args += f' -pr:h {args.profile_host[0]}'
+            for arch in archs:
+                ConanOutput().success(f"Building universal {ref} {arch}")
+                arch_folder = os.path.join(conanfile.build_folder, arch)
+                run(f'conan install --requires={ref}{more_args} -d direct_deploy --deployer-folder "{arch_folder}" -s arch={arch} -b missing', shell=True)
```
memsharded, line 121: This is making Conan re-entrant, and it is not supported, see https://docs.conan.io/2/knowledge/guidelines.html -> Forbidden practices.

> Conan is not re-entrant: Calling the Conan process from Conan itself cannot be done. That includes calling Conan from recipe code, hooks, plugins, and basically every code that already executes when Conan is called. Doing it will result in undefined behavior.

memsharded (commented):
Thanks for your contribution @gmeeker 

Unfortunately, the current approach wouldn't be valid, because of the re-entrancy issue commented below.

I have been thinking of other possible approaches, but at the moment I don't see anything that could be reasonably done without actively modifying recipes and/or build systems integrations that would provide some more built-in approach, because this kind of hacks from the consumer side seem a bit too much of a long shot.

gmeeker:
I realize that with Apple announcing end of support for Intel Macs in 2027, there is probably not a lot of reason to touch Conan's core code. However, I think something like this command (despite its use of Conan internals) could bridge that gap. We've been using a wrapper script that deploys Conan's binaries, runs lipo, and patches the generator files in the top level (executable) recipe's folder. Other users seem to avoid that approach and just lipo two single architecture apps together.
So this is possibly a final attempt to figure out something a little more elegant and integrated. I'm not sure what others will make of it, but it's the best I've come up with.

extensions/commands/install-universal/cmd_install_universal.py
gmeeker: Well, this isn't quite resolved. It can fail due to skipped binaries but I'll give it some thought.

PR #202, opened 2025-07-31
Test/export pkg

danimtb:


danimtb:
This is the result of the build info content of the test that uses the artifactory to gather the files (only conanfile.py and conan_source.tgz found)
image


This is the result in the case that the files are not uploaded to artifactory (upload with --dryrun): Only conanfile.py is found locally
image

danimtb:
Now, this is the result of the same test, with the fix proposed at #201 (now it finds conanfile.py, conan_sources.tgz and conanmanifest.tgz)
image

PR #222, opened 2025-11-25
Add note about CI flow to the build info readme

### LINKED ISSUE #220: [question] Unable to find artifacts, aborting promotion
Hello, I have a question regarding artifact promotion. My scenario:

I follow the steps in https://github.com/conan-io/conan-extensions/blob/main/extensions/commands/art/README.md to create conan packages with build info. They are for different OS and platforms, so I build them distributed and then try to aggregate them later under a single name. This works fine as far as I can tell.

The problem arises when I try to promote my artifact:
```
$ conan art:build-info promote vcd 105 team-conan-dev-local team-conan-local --server artifactory
ERROR: 400: Unable to find artifacts of build 'vcd' #105 from artifactory-build-info repo: aborting promotion.
```

I don't really understand where this error comes from and unfortunately there are not any details about which artifacts could be missing, and I just uploaded everything before in my CI pipeline (I think). I even queried Artifactory for the build info and checked all returned paths in the aggregated build info. All of them exist so it's very unclear which artifacts are supposedly missing.

Can you help me in which direction I would have to investigate or what could be the root cause for this error?

danimtb:
closes #220 
Also included some minor fixes in the readme

### DIFF
--- a/extensions/commands/art/README.md
+++ b/extensions/commands/art/README.md
@@ -43,7 +43,7 @@ conan create . --format json -s build_type=Release > create_release.json
 conan create . --format json -s build_type=Debug > create_debug.json
 ```
 
-**Note**: In case you are using different machines or different Conan cache folders (e.g. when running on a CI system), the recommended flow is to prviously upload the recipe to Artifactory and then use ``conan install --requires=<pkg> --build=<pkg> `` + ``conan test ...`` on each machine to create the binaries for different configurations. This is the way to make sure you operate over the same recipe revision and its metadata files across all the CI machine agents.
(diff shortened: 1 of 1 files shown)


extensions/commands/art/README.md
```diff
@@ -43,6 +43,8 @@ conan create . --format json -s build_type=Release > create_release.json
 conan create . --format json -s build_type=Debug > create_debug.json
 ```
 
+**Note**: In case you are using different machines or different Conan cache folders (e.g. when running on a CI system), the recommended flow is to prviously upload the recipe to Artifactory and then use ``conan install --requires=<pkg> --build=<pkg> `` + ``conan test ...`` on each machine to create the binaries for different configurations. This is the way to make sure you operate over the same recipe revision and its metadata files across all the CI machine agents.
```
czoido:
```suggestion
**Note**: In case you are using different machines or different Conan cache folders (e.g. when running on a CI system), the recommended flow is to previously upload the recipe to Artifactory and then use ``conan install --requires=<pkg> --build=<pkg> `` + ``conan test ...`` on each machine to create the binaries for different configurations. This is the way to make sure you operate over the same recipe revision and its metadata files across all the CI machine agents.
```

PR #223, opened 2025-11-26
[art:build-info] Fix timezone issue on timestamp and regex assert

### LINKED ISSUE #219: `art:build-info` timestamp bug
There's a bug in your `_get_formatted_time()` function in the `art:build-info` command:

https://github.com/conan-io/conan-extensions/blob/76df426353f52af510726646f165fd24aacb5bf3/extensions/commands/art/cmd_build_info.py#L63-L66

I'm no Python expert but it seems to get the time in UTC and then attaches the local TZ offset. This creates wrong timestamps, especially on Windows where the clock often represents UTC and a time zone offset is set.

These two build infos were create on Windows and Linux at the same time but on Artifactory they appear like 1 hour apart:

Windows (wrong):
```
    "started": "2025-11-17T11:34:46.976+0100",
```
Linux (correct):
```
    "started": "2025-11-17T11:33:01.655+0000",
```

---

There's a second bug in the regex that doesn't mask a `.` and thus matches any character, which is probably not intended:
https://github.com/conan-io/conan-extensions/blob/76df426353f52af510726646f165fd24aacb5bf3/extensions/commands/art/cmd_build_info.py#L73

danimtb:
closes #219

### DIFF
--- a/tests/test_build_info.py
+++ b/tests/test_build_info.py
@@ -162,6 +162,7 @@ def test_tool_require_skip_binaries():
 
 
 def test_formatted_time():
+    """Compare local timestamp hours from build-info JSON with current timestamp in UTC"""
     run("conan new cmake_lib -d name=lib1 -d version=1.0")
     run("conan create . -f json > create.json")
 


extensions/commands/art/cmd_build_info.py
```diff
@@ -61,17 +61,17 @@ def _get_hashes(file_path):
...
+    formatted_time = now.strftime(f"%Y-%m-%dT%H:%M:%S.{ms:03d}%z")
 
     # Apparently if the timestamp has the Z the BuildInfo is not correctly identified in Artifactory
     # if local_tz_offset == "+0000":
     #    formatted_time = formatted_time[:-5] + "Z"
 
     # from here: https://github.com/jfrog/build-info-go/blob/9b6f2ec13eedc41ad0f66882e630c2882f90cc76/buildinfo-schema.json#L63
-    if not re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}(Z|[+-]\d{4})$', formatted_time):
-        raise ValueError("Time format does not match BuildInfo required format.")
+    assert re.match(r'^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}(Z|[+-]\d{4})$', formatted_time), \
```
AbrilRBS: Looks good, bit note that upstream has this same bug! https://github.com/jfrog/build-info-go/blob/main/buildinfo-schema.json#L63 shows
`^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.\\d{3}(Z|[+-]\\d{4})$`

czoido (approved):

PR #241, opened 2026-02-17
Add `--continue-on-error` option to `art:build-info promote`

deivse:
Hi, I think this can be very helpful in certain situations. My use case is building on TeamCity with separate build jobs for each OS, and a separate promote job. In such a setup, if a single build job fails, and promote is still ran accidentally, then further promotes will attempt to overwrite existing artifacts unless the account has a delete permission. However, what I really want is to just upload the new stuff and ignore such errors. This allows to treat the promote job as "Promote everything that is new from the latest build job runs" without additional complex logic.

Hopefully this is an acceptable change - it doesn't change the current behaviours, is simple, and can be helpful to some users.

### DIFF
--- a/extensions/commands/art/readme_build_info.md
+++ b/extensions/commands/art/readme_build_info.md
@@ -210,10 +210,13 @@ optional arguments:
 ### ``conan art:build-info promote``
 
 ```
-$ cconan art:build-info promote --help
-usage: conan build-info promote [-h] [-v [V]] [-cc CORE_CONF] [--project PROJECT] [--server SERVER] [--url URL] [--user USER] [--password PASSWORD] [--token TOKEN] [--dependencies]
-                                [--comment COMMENT]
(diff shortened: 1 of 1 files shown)


danimtb:
HI @deivse,

Thanks a lot for opening a PR. Indeed, your use case sounds reasonable, and the implementation is quite easy as it is something that Artifactory API already supports 👍

czoido:
> Tested this without and with the new flag.
> 


@danimtb and is the return code correct (fail return code) in both cases?

deivse:
I've updated the branch. I think we might want to test one last thing if we're being thorough - if the API does return a failure code when _all_ the promotions fail, or if it's still a success. Though to be honest, I really don't have the time to look into it right now, so it's up to you if you wish to test that to ensure the help string for the new argument is 100% correct.

PR #245, opened 2026-03-12
Add test for art:build-info promote --continue-on-error

danimtb:
Follow-up on #241

### DIFF
--- a/tests/test_artifactory_commands.py
+++ b/tests/test_artifactory_commands.py
@@ -401,8 +401,10 @@ def test_build_info_promote_continue_on_error():
     run(f'conan art:build-info upload {build_name}.json --server artifactory')
     # Remove package (not recipe) to force promotion fail
     run("conan remove 'mypkg/1.0:*' -c -r extensions-stg")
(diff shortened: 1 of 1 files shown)


tests/test_artifactory_commands.py
```diff
@@ -380,6 +380,37 @@ def test_build_info_project():
...
+
+    run("conan new cmake_lib -d name=mypkg -d version=1.0 --force")
+    run("conan create . --format json -tf='' > create.json")
+    run("conan upload 'mypkg/1.0' -c -r extensions-stg")
+
+    run(f'conan art:build-info create create.json {build_name} {build_number} extensions-stg --server artifactory > {build_name}.json')
+    run(f'conan art:build-info upload {build_name}.json --server artifactory')
+    # Remove package (not recipe) to force promotion fail
+    run("conan remove 'mypkg/1.0:*' -c -r extensions-stg")
+    run(f'conan art:build-info promote {build_name} {build_number} extensions-stg extensions-prod --server artifactory', error=True)
```
czoido: can we check something on the output here to verify that the thing that fails is what we expect to fail?

PR #253, opened 2026-04-13
[art:build-info] Add python requires as recipe dependencies in create command

danimtb:
closes #251

### DIFF
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -325,8 +325,8 @@ class _BuildInfo:
         for node in nodes.values():
             ref = node.get("ref")
             if ref:
-                transitive_dependencies = node.get("dependencies").keys() if node.get("dependencies").keys() else []
-                python_requires = node.get("python_requires").keys() if node.get("python_requires").keys() else []
+                transitive_dependencies = node.get("dependencies").keys() if node.get("dependencies") else []
+                python_requires = node.get("python_requires").keys() if node.get("python_requires") else []
                 binary = node.get("binary")
 
                 if ref and ((binary == "Build") or (binary in ["Cache", "Download", "Update"] and self._add_cached_deps)):
--- a/extensions/commands/art/cmd_build_info.py
+++ b/extensions/commands/art/cmd_build_info.py
@@ -173,12 +173,17 @@ class _BuildInfo:
             reference = RecipeReference.loads(node.get("ref"))
             return self._conan_api.cache.export_path(reference)
 
-    def get_artifacts(self, node, artifact_type, is_dependency=False):
+
+    def get_artifacts(self, node, artifact_type, is_dependency=False, requested_by=None):
         """
         Function to get artifact information, those artifacts can be added as artifacts of a
         module or as artifacts from dependencies and depending on that the format is
         different. For artifacts of modules they have the keys 'name' and 'path'. If they come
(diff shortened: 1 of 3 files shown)


extensions/commands/art/cmd_build_info.py
```diff
@@ -84,7 +84,7 @@ def _get_requested_by(nodes, node_id, artifact_type):
     requested_by_ids = []
 
     for id, node in nodes["1"].get("dependencies").items():
-        if node.get("direct") == "True":
+        if node.get("direct") is True or node.get("direct") == "True":
```
danimtb: This is a fix, as the requestedBy field was not filled

czoido: I think it's ok to just keep this?
```suggestion
        if node.get("direct") is True:
```

extensions/commands/art/cmd_build_info.py
czoido, line 79: I have been checking and the `requestedBy` field is also adding the artifact itself to its own requester chain. For example, for `base_lib/1.0` (a transitive dependency via `lib/1.0`):


```
"requestedBy": [
    [
        "base_lib/1.0#...",   ← the artifact itself
        "lib/1.0#...",
        "app/1.0#..."
    ]
]
```

I'm not sure this is correct or not, maybe we should check with other types of build infos.

Also, I was thinking that maybe we want to treat the `requestedBy` fix as a separate PR because it may deserve a small refactor.

tests/test_build_info.py
```diff
@@ -161,6 +165,93 @@ def package_info(self):
...
+    graph = json.loads(load("create.json"))["graph"]
+    _fake_conan_sources(graph)
+
+    run("conan art:build-info create create.json build_name 1 repo --with-dependencies > bi.json")
+    build_info = json.loads(load("bi.json"))
+
+    app_recipe_mod = next(
+        m for m in build_info["modules"]
+        if m["id"].startswith("app/1.0") and ":" not in m["id"]
+    )
```
czoido: this is repeated all over the test and the test is a bit difficult to read for me, maybe we can make a dedicated function for these patterns?

extensions/commands/art/cmd_build_info.py
danimtb: ok, let's keep the requestedBy out of the PR 👍

tests/test_build_info.py
danimtb: I will simplify the test

tests/test_build_info.py
```diff
@@ -161,6 +165,62 @@ def package_info(self):
...
+    graph = json.loads(load("create.json"))["graph"]
+    _fake_conan_sources(graph)
+
+    run("conan art:build-info create create.json build_name 1 repo --with-dependencies > bi.json")
+    build_info = json.loads(load("bi.json"))
+
+    modules = build_info["modules"]
+    app_recipe_module = [m for m in modules if m["id"] == "app/1.0#697c25af89f95d7c16b88d97972fb7ad"][0]
+    app_pkg_module = [m for m in modules if
+                      "app/1.0#697c25af89f95d7c16b88d97972fb7ad:19f85b5f0cf7b39158b8bce1a58bcb78449fee9d" in m["id"]][0]
```
czoido, line 201: I think we should avoid hardcoding these so that tests run anywhere?

danimtb: The recipes do not have settings or options, so the recipe revision, PID and package revision are the same on every platform

czoido (approved):

PR #259, opened 2026-07-21
Sketch for size-based cache clean

AbrilRBS:


extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+    "kib": 1024, "mib": 1024 ** 2, "gib": 1024 ** 3, "tib": 1024 ** 4,
+}
+
+# A unit-less number smaller than this is almost certainly a forgotten unit (e.g. '10'
+# meaning 10GB, not 10 bytes). Since such a tiny limit would evict basically the whole
+# cache, we refuse it: clearing the cache is what 'conan remove "*:*"' is for.
+_MIN_UNITLESS_BYTES = 1024 ** 2
+
+
+def _parse_size(text):
```
memsharded: Maybe we want to simplify this, it is not that necessary to be able to support all user inputs, a convention that the input is in Mbs could be fine?

Or if not, I think that understanding Mbs and Gbs is more than enough

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+            pass
+        for pref, _info in prefs.items():
+            try:
+                base = os.path.dirname(conan_api.cache.package_path(pref))
+            except ConanException:
+                continue
+            packages.append({
+                "pref": pref,
+                "size": _folder_size(base),
+                "created": pref.timestamp or 0,
```
memsharded, line 94: Do we want this timestamp (this can be a server timestamp, not the one in the local cache)? When we say "created" it is created in the local cache as a file (installed) or the package revision timestamp?

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+            lines.append(f"Over the size limit, but all {result['protected']} eviction "
+                         f"candidate(s) are protected by --older-than={result['older_than']}; "
+                         f"nothing removed.")
+        else:
+            lines.append("Over the size limit, but there are no package binaries to evict "
+                         "(only recipe/source folders remain). Try 'conan cache clean'.")
+        cli_out_write("\n".join(lines))
+        return
+
+    verb = "Would remove" if result["dry_run"] else "Removed"
```
memsharded: Maybe dry-run is not necessary, this could start with a minimal proposal.

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+        "size_after": result["size_after"],
+        "policy": result["policy"],
+        "older_than": result["older_than"],
+        "dry_run": result["dry_run"],
+        "cancelled": result["cancelled"],
+        "over_limit": result["over_limit"],
+        "protected": result["protected"],
+        "recipes_size": result["recipes_size"],
+        "removed": [{"ref": str(i["pref"]), "size": i["size"]} for i in result["removed"]],
+    }
```
memsharded: ``result.copy()`` + handle ``removed``? so it can be done in 2 lines?

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+    cli_out_write(json.dumps(payload, indent=2))
+
+
+@conan_command(group="Custom commands",
+               formatters={"text": _text_output, "json": _json_output})
+def clean_size(conan_api: ConanAPI, parser, *args):
+    """
+    Prune the local Conan cache down to a maximum size, evicting package binaries by a policy.
+
+    This is the "clean by size" counterpart to the built-in 'conan cache clean' (which only
```
memsharded: Not counterpart, different concept, different files/folders to remove.

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+    wipes source/build/download temp folders). Here, whole package binaries are removed until
+    the cache fits within --max-size. Removal is always explicit (this command), never automatic.
+    """
+    parser.add_argument("--max-size", action=OnceArgument, required=True,
+                        help="Target maximum cache size, e.g. 10GB, 500MiB, 2g. The cache is "
+                             "pruned until it fits within this size. A unit is expected; a bare "
+                             "number is bytes, and a tiny bare number is rejected as a likely typo.")
```
memsharded: Most likely Gbs is the size to go, and forget about parsing. I don't think a cache of <1Gb would be an issue, and in any case, users could input ``--max-size=0.1`` (Gb would be implicit)

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+import re
+import time
+
+from conan.api.conan_api import ConanAPI
+from conan.api.input import UserInput
+from conan.api.model import ListPattern
+from conan.api.output import ConanOutput, cli_out_write
+from conan.cli.command import conan_command, OnceArgument
+from conan.errors import ConanException
+from conan.internal.util.files import human_size
```
memsharded: Internal usage, better avoid

extensions/commands/cache/cmd_clean_size.py
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    if args.dry_run or not to_remove:
+        return result
+
+    # Confirm before deleting (matches 'conan remove'; --confirm / core:non_interactive skip it).
+    total = sum(i["size"] for i in to_remove)
+    message = (f"Remove {len(to_remove)} package binaries ({human_size(total)}) to bring the "
+               f"cache under {human_size(max_size)}?")
+    ui = UserInput(conan_api.config.get("core:non_interactive"))
```
memsharded: We can assume the interactive confirmation is not necessary in this command, and simplify this.

memsharded (approved):
Looking good!

PR #260, opened 2026-08-11
fix(sbom): fixing typo in PURL description

JonasKlamroth:
The recipe rev identifier in the created PURL was misspelled as "rref" instead of "rrev".  For reference this is the correct spelling according to the PURL specification: https://github.com/package-url/purl-spec/blob/f319c8ebe478a9c410d47aa56ed9a42720422562/docs/types/definitions/conan-definition.md?plain=1#L45

### DIFF
--- a/extensions/commands/sbom/cmd_cyclonedx.py
+++ b/extensions/commands/sbom/cmd_cyclonedx.py
@@ -91,7 +91,7 @@ def cyclonedx(conan_api: ConanAPI, parser, *args) -> 'Bom':
             version=node.conanfile.version,
             qualifiers={
                 "prev": node.prev,
-                "rref": node.ref.revision if node.ref else None,
+                "rrev": node.ref.revision if node.ref else None,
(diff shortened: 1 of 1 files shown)


danimtb (approved):

danimtb:
Thanks for the fix!

memsharded:
@danimtb 

The code of the built-in cyclonedx generationis:

```python
def _calculate_bomref(component):
    user = f"&user={component.ref.user}" if component.ref.user else ""
    channel = f"&channel={component.ref.channel}" if component.ref.channel else ""
    return f"pkg:conan/{component.name}@{component.ref.version}?rref={component.ref.revision}{user}{channel}"
```

Does that mean that that ``rref`` is also a typo? should the built-in integration be fixed? where that ``rref`` comes from?

memsharded:
https://github.com/conan-io/conan/issues/20268 changed Conan built-in to ``rrev`` too

PR #262, opened 2026-09-09
art:promote: Promote package metadata

### LINKED ISSUE #250: ``conan art:promote`` might not promote package metadata
The current implementation does a explicit list of files to be promoted.

But there might be package-metadata files to promote, and they will not be promoted. For the ``export/recipe`` part case, it might not be an issue because the full folder is promoted.

Coming from https://github.com/conan-io/conan-extensions/issues/249#issuecomment-4192994745

aagor:
Add promotion of package metadata during `conan art:promote`, too.
Recipe metadata is already promoted.

Fixes #250.


aagor:
Hi @AbrilRBS, thanks for adding a test.

Is there anything I can do to get this forward?

memsharded:
Hi @aagor 

Sorry, we have been busy lately with the conferences CppCon and Roscon, this might still take a bit for final review and decision.

Note that this repo with custom commands might be preferably used from users' forks, to achieve more stability and robustness, as this repo is not released/versioned, so using just the latest head might break things. That way it will also not be urgent to get this merged. Thanks for your patience!

aagor:
Hi @memsharded,

everything's fine.

We do not use this repo directly, but sync it regularly/when we need something.
Given your usual speed (thanks!), I would just have synced this repo to our internal fork ;-).
But no problem at all, I will add it to our internal repo directly and you can take the necessary time :+1:.

